PR #1, opened 2015-10-08
Patch eigen ninja fortran fix

dstoup:
Patch Eigen to fix issues when using the ninja generator.
The current release fails when searching for a Fortran compiler when using ninja.

The patch was to add -G ${CMAKE_GENERATOR} to the COMMAND in execute_process. We would fail on ninja without it, since it would not consider the actual generator used.


kfieldho:
Change the vision-tpl references to fletch references (e.g. vision-tpl_SOURCE_DIR --> fletch_SOURCE_DIR)

PR #7, opened 2015-11-24
Update Boost compiler support for MSVC14 (Visual Studio 2015)

mleotta:
This branch addresses an issue in which Boost is built for Visual Studio 2013 (MSVC12) for any new version of Visual Studio, like 2015.  This patch simply adds a case for the 2015 compiler, but it might be better to add a check for future versions (greater than 20) that produces some sort of error or warning until those versions actually exist.


dstoup:
Handled in another branch

PR #22, opened 2016-01-21
Patch OpenCV to remove INSTALL_NAME_DIR line for modules.

mleotta:
On OS X, the OpenCV libraries have install names prefixed with "lib"
and none of them can be found when using fixup_bundle to produce a
package.  Removing the line that sets INSTALL_NAME_DIR fixes the
problem.


hughed2:
Jenkins test this please

mleotta:
This patch has been accepted upstream into OpenCV https://github.com/opencv/opencv/pull/6006 It is required to generate the installer packages on OS X for MAP-Tk.  Can we please merge it?

PR #24, opened 2016-03-14
patch qt4 to build with vs 2015

billhoffman:
@dstoup here is the patch that builds qt4 with vs 2015


dstoup:
I fixed the vs 2013 issue and added the debug-release option for multi-configuration platforms. That pull request is here:

https://github.com/Kitware/fletch/pull/34

PR #25, opened 2016-03-18
Fix shapelib

judajake:
The download location has moved.  Beta is no longer available.  The updated version does not seem to need the patch (the patch breaks the build)


dstoup:
This patch doesn't build for me. For vision TPL, we used the following URL for the beta and left the patch in place.

http://pkgs.fedoraproject.org/repo/pkgs/shapelib/shapelib-${shapelib_version}.tar.gz/708ea578bc299dcd9f723569d12bee8d/shapelib-${shapelib_version}.tar.gz

dstoup:
Was this branch merged as is? Is should no have been since it doesn't work.

hughed2:
This was part of Juda's add_caffe PR. I'll put in a new PR in a moment to follow vision_tpl's URL and patch.

PR #34, opened 2016-03-30
Qt vs2015

dstoup:
This branch takes the work Bill did, and makes a small fix for vs 2013 and a fix to allow debug-release builds on multi-configuration systems.


dstoup:
It seems like the last commit debug-release, is causing a build issue. I will investigate and either fix, or roll that commit out.

dstoup:
I rolled back the debug-release commit which was causing the build failure. I believe this branch is ready to go, but to be safe, we can wait for the dashboards.

PR #36, opened 2016-04-05
Increased Eigen version from 3.2.7 -> 3.2.8

Purg:
There was an issue in Eigen where it was incorrectly using C++11
features when the compiler did not support them (use of nullptr symbol).
The new version of Eigen has a revised macro that correctly defines
things such that those C++11 features are not used when not actually
supported.

### DIFF
--- a/CMake/fletch-tarballs.cmake
+++ b/CMake/fletch-tarballs.cmake
@@ -40,9 +40,9 @@ set(OpenCV_md5 "32f498451bff1817a60e1aabc2939575")
 list(APPEND fletch_external_sources OpenCV)
 
 # EIGEN
-set(Eigen_version 3.2.7)
+set(Eigen_version 3.2.8)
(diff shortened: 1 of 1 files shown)


linus-sherrill:
Looks good to me.  +1

Purg:
@kfieldho Any reaction opposition to increasing Eigen's version here?

dstoup:
This branch is required for building with CentOS 6. I have tested the failure introduced by 3.2.7 and the fix thanks to this update.

Purg:
@dstoup yay!

mleotta:
@kfieldho , can we land this branch yet?  I just hit this issue again today and had to track it down all over again.  I eventually had that "didn't we already fix this" moment and found my way back to this branch.  It would be nice to merge this so I don't waste time three months from now when I come back to CentOS 5/6 and try to remember again why it doesn't work.

PR #38, opened 2016-04-11
Build full suitesparse

dstoup:
Use the Suitesparse Makefile to build it for platforms which support it.


hughed2:
Jenkins test this please

mleotta:
Work on Linux.  On Mac I get this error:

    [  4%] Performing build step for 'SuiteSparse'
    /bin/sh: line 0: cd: metis-4.0: No such file or directory
    ld: library not found for -lrt
    clang: error: linker command failed with exit code 1 (use -v to see invocation)

Also, we may need some more CMake logic to fallback or default to `BUILD_CXSPARSE_ONLY=OFF`, if LAPACK and OpenBLAS are not found.  These are required if `BUILD_CXSPARSE_ONLY=ON`.  Currently CMake will happily configure Fletch if those requirements are not found and then fail to build SuiteSparse.

mleotta:
Removing the "-lrt" from line 93 of SuiteSparse_config.mk fixes the immediate problem above.  Though I'm not sure if that should be removed in general.

The next problem is that SparseSuite can't find my OpenBlas library.  It's in a non-standard path on my Mac (installed by macports into `/opt/local/lib`).  CMake found it just fine, but it seems that it doesn't pass the found path on to the Make file.

PR #39, opened 2016-04-19
dev/support opencv3 png support

dstoup:
This branch takes the dev/support_opencv3 branch and passes the fletch built PNG to OpenCV. It also fixes a few small issues, e.g. Eigen not actually getting passed to OpenCV.


mleotta:
This helps OpenCV find the right PNG library, but now I getting error about PNG not finding ZLIB because OpenCV is not getting the ZLIB from fletch.

Also I noticed that there is now a `fletch_ENABLE_EIGEN` option that pops up after selecting an OpenCV version, and that is different from `fletch_ENABLE_Eigen`.  And when starting from a clean build there is initial an error from ExternalProject before a version of OpenCV is selected.  Once you select a version that goes away.

CMake/External_OpenCV.cmake
```diff
@@ -1,59 +1,75 @@
...
+endif()
+
+# Include link to contrib repo if enabled
+if (fletch_ENABLE_OpenCV_contrib)
+  set(OpenCV_CONTRIB_ARG "-DOPENCV_EXTRA_MODULES_PATH:PATH=${OpenCV_contrib_MODULE_PATH}")
+  list( APPEND OpenCV_DEPENDS OpenCV_contrib)
 endif()
 
 # Set Eigen dependency if we're locally building it
+option(fletch_ENABLE_EIGEN "Should Eigen Support be turned on for OpenCV?" ${_OpenCV_ENABLE_EIGEN_DEFAULT})
```
mleotta: Ah. I see now that `fletch_ENABLE_EIGEN` is to enable Eigen in OpenCV.  The variable naming makes this very confusing.  I'm not sure why it's not just always on (or on whenever building Eigen), but if it needs to be an option it should be called `fletch_Use_Eigen_in_OpenCV` or something else containing both name Eigen and OpenCV.

mleotta:
There is also an issue where the downloaded files for OpenCV and OpenCV_contrib are poorly named and have the same name so they overwrite each other.  I have a fix for that here mleotta@ff25c55.  I'm not sure if I should pull request yet another variant of this branch.

CMake/External_OpenCV.cmake
dstoup: That variable seems poorly named. I only moved it from the IF(FALSE) block. I don't see any reason to always pass Eigen to OpenCV if built. I will make that change.

dstoup:
Regarding the first issue, that is occurring on the original branch as well, and I believe it has to do with timing of variable initialization surrounding the OpenCV version. I will investigate if we think this branch should take over for the original work. 

Regarding Zlib, that support is added just as easily as PNG. I will update for that as well.

dstoup:
FYI, I have merged my fork back into the original and will address these issues on it. I will update merge requests once these are done

PR #40, opened 2016-04-20
Dev/support opencv3

dstoup:
Add switchable OpenCV version support
Add internal PNG and ZLib support to OpenCV build


mleotta:
you should also set the download name prefix for opencv_contrib as I did in my commit.

mleotta:
also there is a variable with the wrong case causing a Eigen to be disabled (and a build error for me)  Here is the fix:

https://github.com/mleotta/fletch/commit/f9abb33fc8761fa3236ea0b00229db8a4e0a879c

mleotta:
This is not specific to OpenCV 3, but there are still several libraries that OpenCV depends on that are built by fletch but fletch is not telling OpenCV to use the fletch version.  libtiff is the next one that is causing problems for me, but probably not the last.

This is mostly a problem when building fletch on a machine that happens to have system version of many of these libraries installed.

hughed2:
jenkins retest this please

hughed2:
jenkins test this please

CMake/External_OpenCV_contrib.cmake
```diff
@@ -0,0 +1,17 @@
+# The OpenCV contrib repo external project
```
dstoup: This approach is ok, but I think it might be cleaner to add the contents of this file to the External_OpenCV.cmake and run it as a custom step after the OpenCV build is done, e.g. 
  if (BUILD_OpenCV_CONTRIB)
    ExternalProject_Add_Step(OpenCV build_contrib
      COMMENT "Build the OpenCV contrib modules"
      DEPENDEES build
      ........
      )
  endif()
   
or something like that.

mleotta:
In addition to adding OpenCV 3 support, this branch also fixes an issue in which OpenCV was not finding  PNG and ZLIB from fletch and using system packages instead.  This needs to be extended to include TIFF and JPEG.  Every time I build Fletch I have to manually go into the OpenCV build and reconfigure it's CMake variable to point to the Fletch PNG, ZLIB, TIFF, and JPEG.  This is an issue with the current master (OpenCV 2.4) as well.  It may make sense to split off the PNG/ZLIB/TIFF/JPEG fix into a separate branch from the OpenCV 3 support.  Not getting the correct library is a bug that needs to be fixed, while OpenCV 3 is new feature.

I'll point out that this bug is somewhat tricky to test.  It only shows up if you have developer system packages installed for PNG, ZLIB, TIFF, JPEG with different versions from Fletch (many linux machines do).  Also it does not result in compile or link error.  Fletch will build perfectly fine.  The issue shows up for me in external tools that have linked to OpenCV and other Fletch components.  Even there it only shows up at run-time when I am exercising the OpenCV code to load an image and it crashes due to a mismatch in the version of one of these libraries.

CMake/External_OpenCV.cmake
```diff
@@ -42,18 +30,57 @@ if(fletch_DISABLE_GPU_SUPPORT)
...
+
+# PNG
+if(fletch_ENABLE_PNG)
+  list(APPEND OpenCV_EXTRA_BUILD_FLAGS
+    -DPNG_PNG_INCLUDE_DIR=${fletch_PNG_INCLUDE_DIR}
+    -DPNG_LIBRARY_RELEASE=${fletch_PNG_LIBRARY}
+    -DPNG_LIBRARY_Debug=${fletch_PNG_LIBRARY}
+    -DBUILD_PNG=OFF
+    )
+  list(APPEND OpenCV_DEPENDS PNG)
```
dstoup: The PNG args need to contain  -DWITH_PNG=ON

PR #46, opened 2016-06-29
Visual Studio 2015 fixes

hughed2:


dstoup:
First, if these are real bugs, can we also get them upstream? 
Second, it is impossible to see what the actual changes are because the patch is created with a bunch of new files. I generally prefer to have one commit which adds the files to the patch and another commit to apply the changes. That way a reviewer can see the actual changes. Is it possible to format this branch similarly? If not, that's ok, don't want to add extra work, just not going to be easy to provide meaningful feedback otherwise.

hughed2:
Ah! Sorry about formatting, I had it all done like that originally, I reverted a bunch of stuff this morning to get rid of some old redundant commits and kept only the changes themselves. I'll go ahead and update that.

hughed2:
Should be all good now.

hughed2:
This pull request + pull request #40 should now be enough to get a full Visual Studio 2015 fletch build

PR #48, opened 2016-07-11
VTK fix and VXL support

hughed2:


CMake/External_VXL.cmake
```diff
@@ -0,0 +1,128 @@
...
+
+# Not using ffmpeg right now
+#set(VXL_ARGS_VIDL
+#  -DFFMPEG_CONFIG:FILEPATH=IGNORE
+#  -DFFMPEG_INCLUDE1_DIR:PATH=IGNORE
+#  -DFFMPEG_INCLUDE2_DIR:PATH=IGNORE
+#  -DFFMPEG_avcodec_LIBRARY:PATH=IGNORE
+#  -DFFMPEG_avformat_LIBRARY:PATH=IGNORE
+#  -DFFMPEG_avutil_LIBRARY:PATH=IGNORE
+#  -DFFMPEG_swscale_LIBRARY:PATH=IGNORE
```
dstoup: You will need to make a decision regarding FFmpeg in VXL and pass something. If you do nothing, VXL will go rogue. Maybe we need an option to enable FFmpeg in VXL. That is the scenario you have now. Disabling will involve passing IGNORE as you have commented out. Thoughts?

CMake/External_VXL.cmake
```diff
@@ -0,0 +1,128 @@
...
+#  )
+
+if(UNIX)
+  set(VXL_ARGS_V3P
+    -DVXL_USING_NATIVE_ZLIB:BOOL=TRUE
```
dstoup: We have some ZLib stuff up above. We would rather pass our ZLib to VXL if we're building.

CMake/External_VXL.cmake
```diff
@@ -0,0 +1,128 @@
...
+    -DBUILD_CORE_GEOMETRY:BOOL=ON
+    -DBUILD_CORE_NUMERICS:BOOL=ON
+    -DBUILD_CORE_IMAGING:BOOL=ON
+    -DBUILD_CORE_SERIALISATION:BOOL=ON
+    -DBUILD_GEL:BOOL=OFF
+    -DBUILD_MUL:BOOL=OFF
+    -DBUILD_MUL_TOOLS:BOOL=OFF
+    -DBUILD_TBL:BOOL=OFF
+    -DJPEG_LIBRARY:FILEPATH=${JPEG_LIBRARY}
+    -DJPEG_INCLUDE_DIR:PATH=${JPEG_INCLUDE_DIR}
```
dstoup: You need to pass all image formats we build to VXL or you will have runtime mismatches.

CMake/External_VXL.cmake
```diff
@@ -0,0 +1,128 @@
...
+    ${VXL_ARGS_GUI}
+    ${VXL_ARGS_CONTRIB}
+#    ${VXL_ARGS_VIDL}
+    ${VXL_ARGS_V3P}
+    ${VXL_EXTRA_CMAKE_CXX_FLAGS}
+	-DCMAKE_CXX_STANDARD:STRING=11
+    -DBUILD_EXAMPLES:BOOL=OFF
+    -DBUILD_TESTING:BOOL=OFF
+    -DBUILD_DOCUMENTATION:BOOL=OFF
+	-DBUILD_CORE_PROBABILITY:BOOL=ON
```
dstoup: I don't think CORE_PROBABILITY is an option anymore. Make sure you go through what you have here and verify it's required. VXL has moved a lot since we did this stuff.

CMake/External_VXL.cmake
```diff
@@ -0,0 +1,128 @@
...
+    -DBUILD_CORE_GEOMETRY:BOOL=ON
+    -DBUILD_CORE_NUMERICS:BOOL=ON
+    -DBUILD_CORE_IMAGING:BOOL=ON
+    -DBUILD_CORE_SERIALISATION:BOOL=ON
+    -DBUILD_GEL:BOOL=OFF
+    -DBUILD_MUL:BOOL=OFF
+    -DBUILD_MUL_TOOLS:BOOL=OFF
+    -DBUILD_TBL:BOOL=OFF
```
dstoup: We can probably turn off BRL by default, too.

CMake/External_VXL.cmake
```diff
@@ -0,0 +1,128 @@
...
+  )
+
+ExternalProject_Add_Step(VXL forcebuild
+  COMMAND ${CMAKE_COMMAND}
+    -E remove ${fletch_BUILD_PREFIX}/src/VXL-stamp/VXL-build
+  COMMENT "Removing build stamp file for build update (forcebuild)."
+  DEPENDEES configure
+  DEPENDERS build
+  ALWAYS 1
+  )
```
dstoup: What are we doing here ( .. and why )?

Patches/libtiff/Patch.cmake
```diff
@@ -9,3 +9,8 @@ file(COPY
   ${libtiff_patch}/CMakeLists.txt
   DESTINATION ${libtiff_source}
 )
+
+file(COPY
+  ${libtiff_patch}/libtiff/CMakeLists.txt
+  DESTINATION ${libtiff_source}/libtiff
+)
```
dstoup: Need a newline at end of file?

dstoup:
In the future it would have been better to do VXL on a separate PR.

PR #50, opened 2016-07-25
Add a fletch-level option for OpenCV's highgui; defaults to off.

collinsr:
This commit defaults fletch to build OpenCV without the highgui
package. When highgui is enabled, OpenCV searches for ffmpeg, but
(at least on my mac) it finds the "system" (macport) include files
before its local ffmpeg, and fails to build. Turning highgui off
allows OpenCV to build.

As far as I know, no fletch clients require highgui... if this
changes, we can revisit with a more nuanced solution.


hughed2:
I haven't had any issues with that, OpenCV builds fine on Windows/Mac/Linux for me. This is fine for now, but in the medium term, ffmpeg will be packaged in fletch and we can use that for a nicer solution.

PR #51, opened 2016-07-30
Default to building OpenCV's highgui.

collinsr:
MAP-Tk and the OpenCV plugin require highgui. Once we have ffmpeg in fletch,
we can probably remove this option.


collinsr:
@dstoup @mleotta @kfieldho ... I think this branch can be merged? I can't seem to formally assign it or merge it myself, but maybe I'm not managing my github identity correctly?...

PR #52, opened 2016-08-01
Revert "Latest version of VXL is not compatible with VidTK"

mleotta:
The old version of VXL is broken in the way in installs some configured
vnl headers and causes errors building MAP-Tk.  Revert to a newer
version.

This reverts commit 14c5c6b8074bc52781fc6fbbe2fdecf0645250d1.


mleotta:
@hughed2 Here is the change we just talked about.  I think reverting to your previous VXL hash (the newer one) will fix my MAP-Tk build issues.

mleotta:
@dstoup can you merge this?

PR #83, opened 2016-12-15
Fix the problem with installing raw_function.hpp.

dstoup:


mleotta:
works for me.  thanks!

dstoup:
Nice, thanks for testing!

Issue #57, opened 2016-08-22
VXL does not use image libraries from Fletch


mleotta:
When building VXL, it is not configured to build against the image libraries provided by Fletch (e.g. libtiff, libjpeg, libpng, libz).  If the image libraries are provided by Fletch and also by system packages, VXL will prefer the system packages.  This can cause problems with 3rd party projects built against fletch, since they will depend on multiple versions of these image libraries.

mleotta:
Yes, I think it works now

PR #84, opened 2016-12-22
Build VXL against Fletch image libraries

mleotta:
This branch addresses issue #57.  It makes VXL build against the Fletch image libraries (tiff, jpeg, png) instead of system ones.  It does this by telling VXL to search the Fletch install path first.

I have tested this on Linux, but not other platforms.


CMake/External_VXL.cmake
```diff
@@ -111,6 +95,7 @@ ExternalProject_Add(VXL
     -DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}
     -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
     -DCMAKE_INSTALL_LIBDIR:PATH=${fletch_BUILD_INSTALL_PREFIX}/lib
+    -DCMAKE_PREFIX_PATH:PATH=${fletch_BUILD_INSTALL_PREFIX}
```
mleotta: Assuming there are no gotchas that I am not aware of, we might consider adding this line to every project that Fletch is building, or at least the ones that have other dependencies in Fletch.

This is basically telling CMake to prefer finding Fletch packages over system packages.

dstoup: Agreed. We will still need to add each of the supporting packages we build to the DEPENDS list though, so they are built in the correct order.

dstoup:
Code looks good. I will test on Windows in a bit.

PR #88, opened 2017-01-13
Update VXL to latest master hash and checksum

linus-sherrill:


dstoup:
VXL's hash needs to be updated. I merged a fix for the ffmpeg compiler issues. VXL's current master is now correct.

PR #100, opened 2017-02-02
Properly build OpenCV contrib

mleotta:
The key change here is External_OpenCV.cmake:154 the OPENCV_EXTRA_MODULES_PATH was value was not added to the OpenCV_EXTRA_BUILD_FLAGS, so it was downloaded, but not built.  This one line change fixes that.  Of course actually building OpenCV contrib exposes a number of other issues including a cyclical dependency between OpenCV and Caffe.  The easiest solution is to turn off the offending cnn_3dobj module.  The sfm module also using Ceres, but doesn't probably add the include_directories, which is what the patch here is for.


dstoup:
I still don't see anything in build/src/OpenCV_contrib-build when using this branch, is that expected?

mleotta:
Yes, everything still builds in OpenCV-build, there is just more stuff that gets built.  For example, the main thing I need is OpenCV-build/lib/libopencv_xfeatures2d.so.  The build step for OpenCV_contrib is a no-op.  We just need the download step and patch step and then pass the file path to the code into OpenCV when building that.

dstoup:
Thanks, looks good then, confirmed that contrib libraries are built and installed.

PR #113, opened 2017-03-06
Optimize QT building on Windows VS

mattdawkins:


mattdawkins:
Deleting branch, had more commits than I wanted on it.

PR #118, opened 2017-03-21
Adding Windows support for caffe

hughed2:


CMake/External_Caffe.cmake
```diff
@@ -1,17 +1,21 @@
-if (WIN32)
-  # Build option for windows not yet generated
-  message( FATAL_ERROR "Caffe on windows not yet supported" )
-endif()
+#if (WIN32)
+#  # Build option for windows not yet generated
+#  message( FATAL_ERROR "Caffe on windows not yet supported" )
+#endif()
```
linus-sherrill: If this is dead code, could it be removed?

PR #124, opened 2017-04-14
Viame/caffe windows

hughed2:
Adding support for caffe on windows


hughed2:
Lots of extra stuff on here because it was forked off of a branch Dawkins will be merging soon...

PR #128, opened 2017-04-28
Fix boost python build

mleotta:
This are commits cherry-picked from @mattdawkins much larger viame/master branch that specifically address the problem of building Boost.Python on MacOS.


mattdawkins:
Looks good

PR #130, opened 2017-05-15
Fix FFMPEG build for windows

vovythevov:
Packages were no longer available on the FFMPEG website. Instead, the
windows packages were uploaded to data.kiware.com so they are always
accessible.
Only 64bit is supported.


vovythevov:
@dstoup: Would you mind taking a look ?

dstoup:
Looks good. I will give it a build soon, then merge.

PR #136, opened 2017-05-24
Always install FletchConfig in the Fletch_BUILD_INSTALL_PREFIX dir

vovythevov:
Fletch automatically installs everything to the Fletch_BUILD_INSTALL_PREFIX
directory, but kept the cmake configuration file in the build directory.
Because other projects actually build against the install, the FletchConfig.cmake
should be in the install directory too.

CC: @dstoup


hughed2:
Can one of the admins verify this patch?

hughed2:
Can one of the admins verify this patch?

CMakeLists.txt
```diff
@@ -175,7 +176,6 @@ set (COMMON_CMAKE_ARGS
   -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
   -DCMAKE_INSTALL_LIBDIR:PATH=${fletch_BUILD_INSTALL_PREFIX}/lib
   -DCMAKE_PREFIX_PATH:PATH=${fletch_BUILD_INSTALL_PREFIX}
-  -DCMAKE_INSTALL_PREFIX:PATH=${fletch_BUILD_INSTALL_PREFIX}
```
dstoup: This removal is the only change I don't understand/agree with. This directive is important for the subprojects that use CMake. Is it causing some issue?

dstoup: This removal is the only change I don't understand/agree with. This directive is important for the subprojects that use CMake. Is it causing some issue?

vovythevov: @dstoup: That was to remove a duplicate line. I separated my commits to make it more clear. Thanks for the review !

PR #139, opened 2017-06-09
Improve fletch install(s)

vovythevov:
@dstoup: This branch solves some of the issues I have been having, the biggest culprit being having to rebuild if I want to clean the install directory.
I added target so you can:
 - Re-install targets individually ( 35d9e15 )
 - Re-install fletch entirely ( cc5124f )
 - Actually install fletch instead of relying on the "build install". ( 875e5a4 )

Let me know what you think,
Johan

### DIFF
--- a/CMake/External_Boost.cmake
+++ b/CMake/External_Boost.cmake
@@ -61,6 +61,8 @@ ExternalProject_Add(Boost
 )
 add_dependencies(Download Boost-download)
 
+fletch_external_project_force_install(PACKAGE Boost)
+
 set(BOOST_ROOT ${fletch_BUILD_INSTALL_PREFIX})
 
 file(APPEND ${fletch_CONFIG_INPUT} "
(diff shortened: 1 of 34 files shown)


dstoup:
I haven't considered the impact of the branch yet, or tested it thoroughly yet, but there is an initial error during configuration with OpenCV-contrib

1-run cmake
2-select fletch_ENABLE_ALL_PACKAGES
3- error 
CMake Error at CMakeLists.txt:204 (set_target_properties):
   set_target_properties Can not find target to add properties to:
   OpenCV_contrib-install



 CMake Error at CMakeLists.txt:205 (set_target_properties):
   set_target_properties Can not find target to add properties to:
   OpenCV_contrib-install

I will continue on without that package to see what the impact of this branch is on normal processing.

dstoup:
.. more ...
OpecnCV3 build fails on Fedora 23.

vovythevov:
@dstoup: Good catch ! 
I pushed an updated branch that should fix this.

PR #143, opened 2017-06-22
Support MSVC 2017 with OpenCV3

mleotta:
Fletch builds fine without this change, but KWIVER cannot use the OpenCV it produces without this change on Visual Studio 2017.


dstoup:
Jenkins test this please

Patches/OpenCV3/Patch.cmake
```diff
@@ -14,3 +14,11 @@ file(COPY ${OpenCV_patch}/graphcuts.cpp
 file(COPY ${OpenCV_patch}/OpenCVFindLibsVideo.cmake
   DESTINATION ${OpenCV_source}/cmake
 )
+
```
dstoup: Is this patch appropriate for all compilers? There is a lot of code getting patched so it's hard to tell. If we aren't sure, it might be good to use an MSVC version check around the patching itself.

mleotta: I split the this branch into two commits so you can see what the patch is actually changing.  The first commit just copies the files unchanged from the source.  The second makes the patch.  The only change here is a couple of lines in two files to add MSVC 2017 to the list of known Windows compilers, so it should not impact other platforms.  The patch is the same as used upstream.

dstoup: perfect, thanks. I forgot to check the individual commits. I am hoping a dashboard will run, but look unlikely. I will merge in a bit.

PR #144, opened 2017-06-27
Fix FletchConfig.cmake package_DIR variables

mleotta:
This patch fixes numerous issues with the variables exported by Fletch and used by KWIVER to find the Fletch packages.

There are likely other missing variables and variable naming issues (e.g. name case) that need to be fixed, but these fixes are the ones needed to address current build issues, especially on Windows.


dstoup:
Jenkins test this please

PR #145, opened 2017-06-27
Ensure fletchConfig.cmake provides access to VXL and CERES libraries

aaron-bray:


dstoup:
I think this branch is OBE because of the work Matt L did. Let me know if it needs reviewed, otherwise, close.

aaron-bray:
The VXL and KML changes from Matt L. look good
The Ceres change only works on linux, Ceres cmake files install to a
different location on windows
(check out the if block in mine)

The log4cplus changes look fine for now
I have a whole patch of changes to those files and more to get it to
compile and link
currently my log4cplus.dll is not loading properly by the logger manager
loading scheme...
But Linus and I can talk that through

aaron-bray:
Yes, these changes are good

PR #151, opened 2017-07-12
Restored fletch_ROOT variable in exports.

mleotta:
I'm not sure why this was recently removed, but I'm putting it back
because it is the only way to find the fletch install path from projects
using fletch without knowing which fletch components are built.


mleotta:
@vovythevov @dstoup The removal of this line was breaking MAP-Tk builds.  What I really need instead is `fletch_LIBRARY_DIRS` containing the paths to all directories containing the libraries installed by Fletch, but this will do for now.

Most libraries are in `${fletch_ROOT}/lib`, but some (OpenCV on Windows) are in other places, like `${fletch_ROOT}/x86/vc15/Release/bin`.

mleotta:
It seems `fletch_ROOT` has been replaced by `fletch_DIR`, so I will abandon this PR.

PR #155, opened 2017-07-24
Change CUDA disables to enables, point caffe to kitware public

mattdawkins:


dstoup:
Why are there so many commits? Can you rebase -i to squash some of the noise?

CMake/fletch-tarballs.cmake
```diff
@@ -311,15 +311,15 @@ if(NOT WIN32)
...
+  set(Caffe_version "527f97c0692f116ada7cb97eed8172ef7da05416")
+  set(Caffe_url "https://gitlab.kitware.com/kwiver/caffe/repository/archive.zip?ref=${Caffe_version}")
+  set(Caffe_md5 "4ec71f28a797eac7fe3ddcb0fbfab60e")
   list(APPEND fletch_external_sources Caffe)
 else()
-  set(Caffe_version "527f97c0692f116ada7cb97eed8172ef7da05416")
-  set(Caffe_url "https://github.com/BVLC/caffe/archive/${Caffe_version}.zip")
-  set(Caffe_md5 "751e3cd16050e7ecd3981457e41b29cd")
+  set(Caffe_version "7f5cea3b2986a7d2c913b716eb524c27b6b2ba7b")
+  set(Caffe_url "https://gitlab.kitware.com/kwiver/caffe/repository/archive.zip?ref=${Caffe_version}")
```
dstoup: Are we breaking away from the BVLC?

CMake/External_OpenCV.cmake
```diff
@@ -1,7 +1,28 @@
...
+  option(fletch_ENABLE_OpenCV_CUDA "Build OpenCV with CUDA support" TRUE )
+  mark_as_advanced(fletch_ENABLE_OpenCV_CUDA)
+else()
+  set(fletch_ENABLE_OpenCV_CUDA CACHE INTERNAL FALSE)
+endif()
+
+# Allow OpenCV's FFMPEG option to be explicitly turned off
 if(fletch_ENABLE_FFmpeg)
+  option(fletch_ENABLE_OpenCV_FFmpeg "Build OpenCV with FFMPEG support" TRUE )
+  mark_as_advanced(fletch_ENABLE_OpenCV_CUDA)
```
dstoup: Why is the mark_as_advanced shown again, buried in the FFmpeg bits?

CMake/External_OpenCV.cmake
```diff
@@ -197,6 +222,7 @@ ExternalProject_Add(OpenCV
     -DBUILD_SHARED_LIBS:BOOL=True
     -DBUILD_TESTS:BOOL=False
     -DWITH_EIGEN:BOOL=${fletch_ENABLE_EIGEN}
+    -DWITH_JASPER:BOOL=False
```
dstoup: Is there a good reason to ever leave jasper on? What is it about CUDA that messes with the Jasper components?

CMakeLists.txt
```diff
@@ -50,8 +50,18 @@ include(${fletch_CMAKE_DIR}/Utils.cmake)
 add_custom_target(Download)
 add_custom_target(fletch-build-install)
 
-# Option to control GPU support
-option(fletch_DISABLE_GPU_SUPPORT "" FALSE)
+# Options to control GPU support
+option(fletch_BUILD_WITH_CUDA "Build with CUDA support" FALSE)
+option(fletch_BUILD_WITH_CUDNN "Build with CUDNN support" FALSE)
```
dstoup: Is there ever a case for building with CUDNN and not CUDA?

CMake/External_OpenCV.cmake
mattdawkins: I don't know why we would need it turned on and it causes build problems

CMakeLists.txt
mattdawkins: No

CMake/fletch-tarballs.cmake
mattdawkins: We can follow it on our local and have more control in case we need to point caffe to a merger of two branches, as I need on VIAME

CMake/External_OpenCV.cmake
mattdawkins: Fixed

CMake/External_OpenCV.cmake
dstoup: Right, I was suggesting we turn it off in both cases, not just when not enabling CUDA. Not a showstopper though.

mattdawkins: It is disabled in both cases (this is in the main external_project_add for opencv)

PR #160, opened 2017-07-29
[WIP] Add Tensorflow to Build

mattdawkins:
There are still 2 things I know I need to improve on this branch at a minimum

1) Install python packages to correct directories
2) It would be desirable to not have to make a pip package then install it, not very familiar with bazel yet to know if there's a direct install path
3) On windows it would be best if it didn't make alternative rebuilds of protobuf and hdf5


CMake/External_Tensorflow.cmake
```diff
@@ -0,0 +1,90 @@
...
+if(NOT fletch_BUILD_WITH_PYTHON)
+  message(FATAL_ERROR "Error: A build with Python is required for building Tensorflow")
+endif()
+
+if(WIN32)
+  if(NOT fletch_ENABLE_SWIG)
+    message(FATAL_ERROR "Error: SWIG is required for building Tensorflow on Windows")
+  else()
+    set(Tensorflow_DEPENDS ${Tensorflow_DEPENDS} SWIG)
+    if(WIN32)
```
mattdawkins: Redundant, adjust

CMake/External_SWIG.cmake
```diff
@@ -0,0 +1,43 @@
...
+    )
+else()
+  ExternalProject_Add(SWIG
+    URL ${SWIG_url}
+    URL_MD5 ${SWIG_md5}
+    PREFIX ${fletch_BUILD_PREFIX}
+    DOWNLOAD_DIR ${fletch_DOWNLOAD_DIR}
+    INSTALL_DIR ${fletch_BUILD_INSTALL_PREFIX}
+    BUILD_IN_SOURCE 1
+    CONFIGURE_COMMAND ./autogen.sh && ./configure
```
dstoup: Even if you insist `make` is the only way to build on non-win32, there is a macro which specifies this project uses make. You can find that example in another External_* file.

CMake/External_SWIG.cmake
```diff
@@ -0,0 +1,43 @@
...
+
+fletch_external_project_force_install(PACKAGE SWIG)
+
+set(SWIG_ROOT ${fletch_BUILD_INSTALL_PREFIX} CACHE PATH "")
+
+file(APPEND ${fletch_CONFIG_INPUT} "
+#######################################
+# Google SWIG
+#######################################
+set(SWIG_ROOT @SWIG_ROOT@)
```
dstoup: There is a new pattern in fletch for specifying the ROOTs. Please look at something like ZLib and copy what is there.

CMake/External_SWIG.cmake
```diff
@@ -0,0 +1,43 @@
+
+if(WIN32)
```
dstoup: Isn't python required for SWIG? You have a if(NOT fletch_BUILD_WITH_PYTHON) block in the tensorflow CMake file, but not here.

CMake/External_Tensorflow.cmake
```diff
@@ -0,0 +1,90 @@
...
+
+if(NOT fletch_BUILD_WITH_PYTHON)
+  message(FATAL_ERROR "Error: A build with Python is required for building Tensorflow")
+endif()
+
+if(WIN32)
+  if(NOT fletch_ENABLE_SWIG)
+    message(FATAL_ERROR "Error: SWIG is required for building Tensorflow on Windows")
+  else()
+    set(Tensorflow_DEPENDS ${Tensorflow_DEPENDS} SWIG)
```
dstoup: Why are you not requiring Enable_SWIG on Iinux?

CMake/External_Tensorflow.cmake
```diff
@@ -0,0 +1,90 @@
...
+    if(WIN32)
+      set(SWIG_BUILD_FLAGS
+        -DSWIG_DIR=${fletch_BUILD_PREFIX}/src/SWIG-build
+        -DSWIG_EXECUTABLE=${fletch_BUILD_PREFIX}/src/SWIG-build/swig.exe)
```
dstoup: Use SWIG_ROOT here for generic access. Don't use the one in the swig build directory.

CMake/External_Tensorflow.cmake
```diff
@@ -0,0 +1,90 @@
...
+
+  ExternalProject_Add(Tensorflow
+    DEPENDS ${Tensorflow_DEPENDS}
+    URL ${Tensorflow_url}
+    URL_MD5 ${Tensorflow_md5}
+    PREFIX ${fletch_BUILD_PREFIX}
+    DOWNLOAD_DIR ${fletch_DOWNLOAD_DIR}
+    INSTALL_DIR ${fletch_BUILD_INSTALL_PREFIX}
+    BUILD_IN_SOURCE 1
+    CONFIGURE_COMMAND ./configure
```
dstoup: Use CMake for linux build too, if possible.

CMake/External_Tensorflow.cmake
```diff
@@ -0,0 +1,90 @@
...
+
+fletch_external_project_force_install(PACKAGE Tensorflow)
+
+set(Tensorflow_ROOT ${fletch_BUILD_INSTALL_PREFIX} CACHE PATH "" FORCE)
+
+file(APPEND ${fletch_CONFIG_INPUT} "
+########################################
+# Tensorflow
+########################################
+set(Tensorflow_ROOT @Tensorflow_ROOT@)
```
dstoup: See comment in SWIG file about new pattern here.

CMake/fletch-tarballs.cmake
```diff
@@ -286,6 +286,19 @@ set(HDF5_url "https://support.hdfgroup.org/ftp/HDF5/prev-releases/hdf5-${HDF5_ma
 set(HDF5_md5 "bdf0fc3d648679eeb5d7b4b78f92a83f")
 list(APPEND fletch_external_sources HDF5)
 
+# SWIG
+if(WIN32)
+  set(SWIG_version "cmake")
+  set(SWIG_url "https://github.com/swig/swig/archive/${SWIG_version}.zip")
+  set(SWIG_md5 "b691d887f8919e6690c2a124f80628bf")
+  list(APPEND fletch_external_sources SWIG)
+else()
```
dstoup: Why not use CMake version for both?

CMakeLists.txt
```diff
@@ -62,6 +62,9 @@ if (fletch_BUILD_WITH_CUDA)
     set( CUDNN_ROOT "" CACHE PATH "CUDNN root folder" )
     mark_as_advanced( CUDNN_ROOT )
   endif()
+
+  option(fletch_BUILD_CUDA_CSTD98 "Build all cuda code with C++ standard 98" FALSE)
```
dstoup: Why build the Cuda code differently from other code? Is it a problem to enable?disable C++ 11 for the entire build?

dstoup (changes requested):

CMake/External_Caffe.cmake
```diff
@@ -255,6 +255,13 @@ else()
     )
 endif()
 
+if(fletch_BUILD_CUDA_CSTD98)
```
mattdawkins: I changed it to FORCE_CUDA_CSTD98 in an upcoming commit, as it is here, it's still an advanced flag. It was required to build it in this fashion on my home Fedora 24 PC.

CMake/External_Caffe.cmake
```diff
@@ -255,6 +255,13 @@ else()
...
+  set( CAFFE_CXXFLAGS_OVERRIDE
+    -DCMAKE_CXX_FLAGS=-std=c++98 )
```
mattdawkins: The point isn't to specify it for the entire repository, only projects using CUDA.

CMake/External_SWIG.cmake
mattdawkins: I concur with your assessment. Tensorflow (latest), however, recommends using bazel on linux and cmake on windows.

CMake/External_Tensorflow.cmake
mattdawkins: It's only required for the cmake build

CMake/External_Tensorflow.cmake
mattdawkins: See above.

PR #167, opened 2017-08-03
Update to OpenBLAS-0.2.20 

Erotemic:
### Overview

I updated OpenBLAS to version 0.2.20, built it, and doesn't seem to have broken anything. When doing the update, I versioned the code in the Patches directory so it is easier to dynamically upgrade and downgrade the library. 

-----------

### Issue

When trying to build fletch, I ran into an error on OpenBLAS. OpenBLAS 0.2.15 seemed to not recognize the architecture of my Intel i7-7700HQ CPU (mode=158) and produced this error

```bash
[ 12%] Performing build step for 'OpenBLAS'
getarch_2nd.c: In function ‘main’:
getarch_2nd.c:12:35: error: ‘SGEMM_DEFAULT_UNROLL_M’ undeclared (first use in this function)
     printf("SGEMM_UNROLL_M=%d\n", SGEMM_DEFAULT_UNROLL_M);
                                   ^
...
                                                  ^
make[3]: *** [getarch_2nd] Error 1
Makefile:133: *** OpenBLAS: Detecting CPU failed. Please set TARGET explicitly, e.g. make TARGET=your_cpu_target. Please read README for the detail..  Stop.
CMakeFiles/OpenBLAS.dir/build.make:113: recipe for target 'build/src/OpenBLAS-stamp/OpenBLAS-build' failed
make[2]: *** [build/src/OpenBLAS-stamp/OpenBLAS-build] Error 2
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/OpenBLAS.dir/all' failed
make[1]: *** [CMakeFiles/OpenBLAS.dir/all] Error 2
```

This is because `gcc -march=native -Q --help=target|grep march` specified my architecture as BROADWELL, which I believe should be mapped to HASWELL in OpenBLAS. 

-------

### Fix1 - Manually editing the make command 

The first fix for this was to simply specify the target when I made fletch as so:

```bash
make TARGET=HASWELL
```

This is cumbersome and ugly, so I attempted to find another workaround. 

### Fix2 - Updating

I noticed that this issue didn't happen on the latest version of OpenBLAS (v0.2.20), so I updated fletch to target that new version. I didn't want to completely drop the old version, so I modified fletch-tarballs.cmake to choose the url based on a cmake variable (currently hidden from the user). 

When doing so I found that the patch applied to 0.2.15 was merged into OpenBLAS's master in 0.2.16. Because I was upgrading to 0.2.20, that patch should already be applied and there is no reason it would be needed unless I switched back to 0.2.15. Considering this, I moved the patch into a folder reflecting the version that it should be applied to. 

I think it would be a good idea to continue versioning the patches in the future when upgrading other projects. I was doing some experiments with newer versions of VTK and OpenCV and I found this naming scheme useful in quickly testing new versions and reverting if they didn't work. 

#### Versioning Patches

The idea is that all patches should go in a subfolder `Patches/<libname>/<version>` , where `<libname>` is the name of the package and `<version>` is the string defined in the `fletch-tarballs.cmake` file. 
Then, in `External_<libname>.cmake`, you check if the subfolder exists. If it does you make a patch command and apply it, otherwise there is no patch.

For example this is what I did with OpenBLAS:

```cmake
  # Add a patch if one exists for the requested version
  set(OpenBLAS_patch ${fletch_SOURCE_DIR}/Patches/OpenBLAS/${OpenBLAS_version})
  if (EXISTS ${OpenBLAS_patch})
    set(OpenBLAS_PATCH_COMMAND ${CMAKE_COMMAND}
      -DOpenBLAS_patch:PATH=${OpenBLAS_patch}
      -DOpenBLAS_source:PATH=${fletch_BUILD_PREFIX}/src/OpenBLAS
      -P ${OpenBLAS_patch}/Patch.cmake
      )
  elseif()
    set(OpenBLAS_PATCH_COMMAND "")
  endif()
```

Then when calling `ExternalProject_Add`, you simply set `PATCH_COMMAND ${OpenBLAS_PATCH_COMMAND}`

Even, if we don't support dynamically changing versions of the libraries, I think this will still be helpful in keeping the patches organized. I also don't think its necessary to do this to every external library right now, but it might be a good idea to move to this scheme whenever a package is next updated.


Erotemic:
Closing in favor of #173

PR #168, opened 2017-08-03
Added explicit numpy for caffe

Erotemic:
When trying to build Caffe with fletch, it crashed on an error saying that it could not find numpy. I added an explicit `find_package` for NumPy and passed the python variables down to Caffe. This seems to fix the issue. 

Side note: The FindNumpy script that I added is the same one used in Caffe. I find it odd that Caffe wasn't able to find it by itself considering its using the same script. My current guess it that it might have to do with the fact that I installed numpy to my python2 virtual environment.


Erotemic:
closing and reopening with correct branch naming conventions

PR #170, opened 2017-08-04
Create v1.1.0 release notes

mleotta:
This is mostly to put the file in place.  @mattdawkins and @hughed2 should probably update my description of what already has landed in master since v1.0.0.


dstoup:
I will can the git history and get make this file current. Going forward, Release Notes updates are an expected part of a PR.

PR #171, opened 2017-08-04
Upgrade Ceres Solver

mleotta:
Upgrading the version of Ceres Solver seem to build fine in Fletch, but it still needs more testing in KWIVER before we merge.


dstoup:
Jenkins test this please

mleotta:
It seems I may have jumped the gun on this one.  As @hughed2 discovered, the upgraded Ceres builds fine, but there are some issues with this version's CMake exports so KWIVER has trouble finding it.  I'll make some patches for Fletch and also get them upstream.

PR #174, opened 2017-08-04
Versioned OpenCV and added patch for 3.1.0

Erotemic:
This PR updates OpenCV patches to use the versioning conventions proposed in #173 
This exposes 3.3.0 as an option, but leaves the current version (3.1.0) as the default.  

In addition, this patches a bug in version 3.1.0 (which was fixed by 3.3.0), which is described [here](http://answers.opencv.org/question/86541/opencv-make/). This bug would cause the build to fail if the string `cuda` was found anywhere in the path to the build directory.


Erotemic:
I've changed and rebased this PR a bit since its original submission. 

By default OpenCV no longer tries to link to VTK now. 

3.2 is now no longer exposed. Only 3.3 (the new version) and the existing (3.1.0 and 2.4.13) are now exposed. By default it targets the stable 3.1.0. 

When 3.3.0 is turned on I've added a patch that fixes a cxx11 flag in a cmake file.

dstoup:
A few  things I need to get my head around here first, then I will review.

1 - Is the idea of adding 3.2.0 while keeping 3.1.0 temporary while we evaluate the capabilities of the newer version? If that's the case, then ok. If that isn't the case, then I am against adding addition versions.  

2 - I'm not sure how I feel about adding more package specific options for enabling VTK for example. We have avoided that so far because the  combinatorics of package interactions in fletch is huge and if we start adding flags to enable all of them, fletch config will become too unwieldy. Is there a good reason to enable/disable VTK support in OpenCV specifically? In other words, is there any danger of always enabling when VTK is built, or always disabling since I don't think anyone here is using it. If we decide to go down the road of allowing users to select which supporting packages to enable/disable, then I think we need some engineering around it to group the options by package or something similar.

The above are just my thoughts. It would be good to discuss with a wider audience.

CMake/External_OpenCV.cmake
```diff
@@ -21,6 +24,23 @@ else()
   set(fletch_ENABLE_OpenCV_FFmpeg CACHE INTERNAL FALSE)
 endif()
 
+# Allow OpenCV's VTK option to be explicitly turned off
+if(fletch_ENABLE_VTK)
+  # By default turn opencv-VTK support off, even when building with VTK
+  option(fletch_ENABLE_OpenCV_VTK "Build OpenCV with VTK support" FALSE )
+  mark_as_advanced(fletch_ENABLE_OpenCV_VTK)
+else()
+  set(fletch_ENABLE_OpenCV_VTK CACHE INTERNAL FALSE)
```
dstoup: I think if we user CACHE INTERNAL FALSE here, we lose the ability to reenable it with the same cache. I believe we had the same issue with Darknet. I believe @aaron-bray came up with a decent working paradigm for this type of action. Hopefully he will chime in.

CMake/External_OpenCV.cmake
```diff
@@ -41,6 +61,20 @@ else()
...
 
+# Set VTK dependency if we're locally building it
+if(fletch_ENABLE_OpenCV_VTK)
+  message(STATUS "OpenCV depending on fletch VTK")
+  if (OpenCV_version VERSION_LESS 3.2.0 AND VTK_version VERSION_GREATER_EQUAL 7.0)
+    message(FATAL_ERRROR "OpenCV versions before 3.2 can only handle pre 7.0 VTK versions")
+  endif()
+  list(APPEND OpenCV_DEPENDS VTK)
+  list(APPEND OpenCV_EXTRA_BUILD_FLAGS -DWITH_VTK:BOOL=TRUE
+    -DVTK_DIR:PATH=${fletch_BUILD_INSTALL_PREFIX}/lib/cmake/vtk-${VTK_version}
```
dstoup: We have the VTK_DIR variable exposed in the External_VTK.cmake file. Is that variable valid at this point in the configuration? If so, we should use it.

CMake/External_OpenCV.cmake
```diff
@@ -53,6 +87,7 @@ else()
   list(APPEND OpenCV_EXTRA_BUILD_FLAGS -DWITH_EIGEN:BOOL=FALSE)
```
dstoup: Perhaps I don't understand the OpenCV configuration, but shouldn't WITH_EIGEN=TRUE if we've enabled Eigen? I realize this code is existing, but would be good to make sure it's correct.

CMake/External_OpenCV.cmake
```diff
@@ -173,16 +208,17 @@ else()
   list(APPEND OpenCV_EXTRA_BUILD_FLAGS -DBUILD_JPEG=ON)
 endif()
 
-if (OpenCV_SELECT_VERSION VERSION_EQUAL 2.4.11)
+
+# If a patch file exists for this version, apply it
+set (OpenCV_patch ${fletch_SOURCE_DIR}/Patches/OpenCV/${OpenCV_SELECT_VERSION})
```
dstoup: Great change ... thank you!

CMake/External_OpenCV.cmake
```diff
@@ -173,16 +208,17 @@ else()
...
     -DOpenCV_source:PATH=${fletch_BUILD_PREFIX}/src/OpenCV
-    -P ${fletch_SOURCE_DIR}/Patches/OpenCV2/Patch.cmake)
+    -P ${OpenCV_patch}/Patch.cmake
+    )
 else()
-  set(OPENCV_PATCH_COMMAND ${CMAKE_COMMAND}
-    -DOpenCV_patch:PATH=${fletch_SOURCE_DIR}/Patches/OpenCV3
-    -DOpenCV_source:PATH=${fletch_BUILD_PREFIX}/src/OpenCV
-    -P ${fletch_SOURCE_DIR}/Patches/OpenCV3/Patch.cmake)
+  set(OPENCV_PATCH_COMMAND "")
```
dstoup: Don't think we need this else.

CMake/External_OpenCV.cmake
```diff
@@ -200,6 +236,11 @@ if (CMAKE_COMPILER_IS_GNUCC AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 6)
   list(APPEND OpenCV_EXTRA_BUILD_FLAGS -DENABLE_PRECOMPILED_HEADERS:BOOL=OFF)
 endif()
 
+# OpenCV 3.3.0 has an option to enable C++ 11
+if (fletch_BUILD_CXX11)
+  list(APPEND OpenCV_EXTRA_BUILD_FLAGS -DENABLE_CXX11:BOOL=ON)
```
dstoup: Doesn't OpenCV work with the CMAKE_CXX_STANDARD:STRING=11? We already pass that information into the configuration. If it doesn't then this is fine. If it does, it would be good to use the internal CMake way.

CMake/External_OpenCV_contrib.cmake
```diff
@@ -1,7 +1,14 @@
...
-  -P ${fletch_SOURCE_DIR}/Patches/OpenCV_contrib/Patch.cmake)
+# If a patch file exists for this version, apply it
+set (OpenCV_contrib_patch ${fletch_SOURCE_DIR}/Patches/OpenCV_contrib/${OpenCV_SELECT_VERSION})
+if (EXISTS ${OpenCV_contrib_patch})
+  set(OPENCV_CONTRIB_PATCH_COMMAND ${CMAKE_COMMAND}
+    -DOpenCV_contrib_patch:PATH=${OpenCV_contrib_patch}
+    -DOpenCV_contrib_source:PATH=${fletch_BUILD_PREFIX}/src/OpenCV_contrib
+    -P ${OpenCV_contrib_patch}/Patch.cmake)
+else()
+  set(OPENCV_CONTRIB_PATCH_COMMAND "")
```
dstoup: See above ....

CMake/External_OpenCV.cmake
dstoup: Edit: Having said the above, is this else here in case the user switches OpenCV versions?

CMake/fletch-tarballs.cmake
```diff
@@ -105,13 +105,13 @@ set(Eigen_dlname "eigen-${Eigen_version}.tar.gz")
...
+# Support 2.4.13 and 3.1, and 3.3 optionally
 if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES)
-  set(OpenCV_SELECT_VERSION 3.1.0 CACHE STRING "Select the  version of OpenCV to build.")
-  set_property(CACHE OpenCV_SELECT_VERSION PROPERTY STRINGS 2.4.11 3.1.0)
+  set(OpenCV_SELECT_VERSION 3.3.0 CACHE STRING "Select the  version of OpenCV to build.")
+  set_property(CACHE OpenCV_SELECT_VERSION PROPERTY STRINGS 2.4.13 3.1.0 3.3.0)
 
   # Expose optional contrib repo when enabling OpenCV version >= 3.x
-  if (OpenCV_SELECT_VERSION VERSION_EQUAL 3.0.0 OR OpenCV_SELECT_VERSION VERSION_GREATER 3.0.0)
+  if (OpenCV_SELECT_VERSION VERSION_GREATER_EQUAL 3.0.0)
```
dstoup: VERSION_GREATER_EQUAL is a very new CMake construct. It is not supported in versions fletch claims to support. I don't know when it came in, but likely it is in the 3.8 range. Sorry, by we need to use the clunkier
    # Just picking some version above what we support. Not less than 2.5 means version 3.0 or greater :(
    if (NOT OpenCV_SELECT_VERSION VERSION_LESS 2.5)

CMake/fletch-tarballs.cmake
```diff
@@ -121,24 +121,30 @@ if ( NOT fletch_ENABLE_OpenCV OR OpenCV_SELECT_VERSION VERSION_LESS 3.0.0 )
   unset(fletch_ENABLE_OpenCV_contrib CACHE)
 endif()
 
-if (OpenCV_SELECT_VERSION VERSION_EQUAL 3.1.0)
-  set(OpenCV_version "3.1.0")
+if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES)
+  set(OpenCV_version ${OpenCV_SELECT_VERSION})
```
dstoup: This value is already set above in the same if block. Is there a way to cleanly add the following code into the above  
   if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES)
block?

CMake/fletch-tarballs.cmake
```diff
@@ -121,24 +121,30 @@ if ( NOT fletch_ENABLE_OpenCV OR OpenCV_SELECT_VERSION VERSION_LESS 3.0.0 )
...
   set(OpenCV_md5 "6082ee2124d4066581a7386972bfd52a")
-  # Paired contrib repo information
-  set(OpenCV_contrib_version "${OpenCV_version}")
-  set(OpenCV_contrib_url "http://github.com/Itseez/opencv_contrib/archive/${OpenCV_contrib_version}.zip")
   set(OpenCV_contrib_md5 "0d0bfeabe539542791b465ec1c7c90e6")
-  set(OpenCV_contrib_dlname "opencv-contrib-${OpenCV_version}.zip")
-elseif (OpenCV_SELECT_VERSION VERSION_EQUAL 2.4.11)
-  set(OpenCV_version "2.4.13")
-  set(OpenCV_url "http://github.com/Itseez/opencv/archive/${OpenCV_version}.zip")
+elseif (OpenCV_version VERSION_EQUAL 2.4.13)
```
dstoup: I hate to make more work, but since we are bumping the minor of version 2, it would be good to see if we need the Patches we currently have for it. It's possible that some of them have been fixed upstream. You could try a quick build of 2 without applying them just to see.  Ideally we will use code as 'patch-free' as possible.

Patches/OpenCV/3.1.0/common.cmake
```diff
@@ -0,0 +1,150 @@
+# This file is included from a subdirectory
```
dstoup: Again, not trying to make more work but, my preference for patching code is to do it in 2 commits. The first commit adds the file, the second commit applies the patch. That allows us to see exactly what was changed in the file.

Having said that, I think that line should be added to a developer contribution document if we have such a thing.

CMake/External_OpenCV.cmake
aaron-bray: Change the ALL the set calls with INTERNAL to an unset call
such as:
unset(fletch_ENABLE_OpenCV_VTK CACHE)

https://github.com/Kitware/fletch/pull/166/commits/083780f00d8c1c77d9bb329a85d3c594fa26d3c0

Erotemic:
Sorry, the original PR text is a bit old. Let me clarify. 

The idea of this patch and every other "versioning" patch I've submitted (except for OpenBLAS) is to: 
* to expose an experimental version while still defaulting to the current stable version and
* to put patches in sub-directories that reflect the version they those patches are targeted towards. 

So, yes, your assumptions on point 1 are correct. This is only mean to be a first step in a very gentle transition to newer software versions. I'm envisioning that there will only ever be at most 2 versions of each software package as an option (maybe 3 in the case of OpenCV). The current stable version, and the version we are working towards. 

On your second point, I think just disabling OpenCV VTK is perfectly reasonable.

CMake/External_OpenCV.cmake
Erotemic: Yes, there is defiantly a problem with this. It exists for the other flags like `fletch_ENABLE_OpenCV_CUDA` and `fletch_ENABLE_OpenCV_FFmpeg` as well. I originally had rewrote all of them to attempt to fix the issue, but I thought my solution was too complex, so I just removed it and followed the pattern of the other variables. 

For reference my solution was to make a special `advanced_option` function that was only visible if a condition was met. 

```cmake
function(advanced_option varname docstring default is_relevant)
  # Allows advanced options to be shown in the gui only when appropriate
  # Once an variable is set as CACHE INTERAL, it seems the only way to make it
  # visible in the gui again is to unset it.
  if (DEFINED ${varname})
    # Remember the previous value of the variable
    set(value ${${varname}})
    unset(${varname} CACHE)
  else()
    # Use the default on the first call or if the var has been externally unset
    set(value ${default})
  endif()

  if (${is_relevant})
    # Show the last known value of the variable, but only in advanced mode
    option(${varname} ${docstring} ${value})
    mark_as_advanced(${varname})
  else()
    # Keep the variable around in internal cache, but dont show it even in
    # advanced mode.
    set(${varname} ${value} CACHE INTERNAL ${docstring})
  endif()
endfunction()
```

CMake/External_OpenCV.cmake
Erotemic: In the current state of the PR turning on VTK will actually break because External_OpenCV is included before External_VTK. Originally I re-arranged the order in fletch-tarballs, but that introduced a lot of merge conflicts in my upstream so I reverted for the time being.

I'm currently leaning towards the idea of removing the option to enable VTK support for OpenCV, but I think its still a good idea to keep this logic here in case we do need to enable it in the future. That being said, I'll change to use the VTK_DIR. We'd need to make that ordering change anyway.

CMake/External_OpenCV.cmake
Erotemic: Oops, that's a copy paste error. That should read `WITH_VTK:BOOL=FALSE`

CMake/External_OpenCV.cmake
aaron-bray: yeah, but once you set a variable to INTERNAL, it never shows up again in the GUI
So if you go back and try to turn it back on via the GUI, you can't...

Yeah, CUDA and FFmpeg need to be updated on this PR or mine, yours if pushing forward, so I vote you do it here. I can cherry pick another PR with Dave when he gets to mine if need be.

CMake/External_OpenCV.cmake
Erotemic: The reason its there is because my `CMake` skills are intermediate and I didn't know if an undefined variable threw an error or evaluated to an empty string. I tend to err on the side of including an `else` cases when I code, so this is a result of that and my laziness for not checking the thing I wasn't sure about.

However, now that you mention the second point, it probably should be there to clobber the value of `OPENCV_PATCH_COMMAND` if the versions switch. So, I'll amend my previous statement and say I wrote that code for very conscious and good reasons (I'll add a comment explaining the behavior).

CMake/External_OpenCV.cmake
Erotemic: This is a new flag in the OpenCV 3.3.0 `CMakeLists.txt`, and it wont build properly without it.

CMake/fletch-tarballs.cmake
Erotemic: Reverting. I prefer to use 3.x instead of an arbitrary version because the code is more readable even it is more clunky.

CMake/fletch-tarballs.cmake
Erotemic: This can be done.

CMake/fletch-tarballs.cmake
Erotemic: Hmm, I didn't think I bumped the minor version of `2.4`. 

Looking at only the context of the diff isn't this fixing a typo? We were always using 2.4.13, but the `OpenCV_SELECT_VERSION` was just switching on the value `2.4.11` for whatever reason.

CMake/fletch-tarballs.cmake
Erotemic: ...err, actually it seems like the logic for disabling `fletch_ENABLE_OpenCV_contrib` based on opencv settings makes this less trivial change. Maybe I can rework it.

Erotemic: I think I can move it all into the same block, but I have to duplicate a `unset(fletch_ENABLE_OpenCV_contrib CACHE)` line. Overall I think it makes it slightly cleaner.

CMake/External_OpenCV.cmake
Erotemic: Actually, now that I'm thinking about it, is the state of the variable persistent (my guess is no since it is not cached)? In that case  maybe the else is unnecessary. I'll test it.

CMake/External_OpenCV.cmake
dstoup: @Erotemic , It does exist for others, yes. The FFmpeg one always bothered me and I think the CUDA one was needed because it would break the build under certain uncontrollable conditions, like CUDA version. I don't know if I care if it remains for a while because of build related issues, but it we are able to, as a group, decide if it should be on/off, then that is better.

CMake/External_OpenCV.cmake
dstoup: At this point, either way is ok. It will be good to test whether we can just enable it when VTK is enabled and be done with it. The catch is, will something in  OpenCV break as a result. The only other downside would be a probable increase in build time.

CMake/External_OpenCV.cmake
Erotemic: I can do it here.

CMake/External_OpenCV.cmake
dstoup: That make more sense :)

CMake/External_OpenCV.cmake
dstoup: Sounds good. If that is the case, it still might be good to leave it with a comment. That makes the reasoning behind having it clear and explicit and protects from a future change to this logic, long after the memories of why are gone:)

CMake/External_OpenCV.cmake
dstoup: Great catch and bad news for us ...  I added that variable as the overall control for CXX_11 in fletch. That really leaves us 2 choices, 1) increase our minimum_cmake_version to 3.3.0 or 2) replace the current CXX_11 control with something less elegant.

I rather prefer the first since we have already discussed raising the CMake version and this gives us a stake in the ground about why we selected the version we did. That said, it is a larger issue which needs more buyin. Leave it for now and I will add an 'issue' to remember the current problem.

dstoup: 'Issue' reported.

CMake/fletch-tarballs.cmake
dstoup: That is good. You can say NOT VERSION_LESS 3.0.0 since our starting OpenCV version is greater than that .. much better idea!

CMake/External_OpenCV.cmake
aaron-bray: Note only MSVC 2015 and 2017 fully support C++11
MSVC 2013 has partial support, but not all, will probably need to have a check somewhere if we still want to support 2013

here is a list of popular compilers and their initial version to implement all of C++11 :

GCC 4.8.1 and later
Clang 3.3 and later
MSVC 2015 and later

CMake/fletch-tarballs.cmake
dstoup: If it's a big rework, skip for now, it's ok to have the second block if that is the cleaner option.

CMake/fletch-tarballs.cmake
dstoup: Ah ... you are probablyl right.  I seem to recall that, but never got to fixing it. Thanks, disregard and carry on :)

dstoup:
Thanks for the clarification, that is helpful for reviewing all of the version branches. That was my guess, but good to be explicit.

The release-notes is going to be a bit of a pain. I will need to be a lot more methodical in merging branches. That said, I wonder if we should add major sections to bottom part to help minimize trouble. What I just thought of is, we could have package level changes, like OpenCV, PROJ, General .. etc, then if you are making bug fixes to OpenCV for example, they would go into that section and should be less likely to conflict with bug fixes in other packages or with a general fletch level change, etc. @mleotta do you have any suggestions or wisdom about this issue?

CMake/External_OpenCV_contrib.cmake
Erotemic: I tested this and it is not needed. The variable state isn't persistent.

CMake/External_OpenCV.cmake
Erotemic: I just confirmed that OpenCV 3.3.0 does can build without C++11. 

The issue is if you enable C++11 in fletch but dont set OpenCV's `ENABLE_CXX11` flag, you will get build errors. I'm working on confirming what these are.

Erotemic: Hmm, I can't seem to reproduce the error (using a minimal first effort). I know I came across a C++11 bug at some point when building. Maybe it only happens when other build settings are on, but perhaps it was a fluke.

I can either keep it, or I can remove it for now until I can reproduce and document what that issue was. Let me know which one.

Patches/OpenCV/3.1.0/common.cmake
Erotemic: I added a commit that reverts to the original version and then another one that applies the patch. Hopefully that works. (Isn't all of this going to get squashed on merge anyway)?

CMake/fletch-tarballs.cmake
Erotemic: Fixed this.

CMake/fletch-tarballs.cmake
Erotemic: Its already in,  see if you like it.

CMake/External_OpenCV.cmake
dstoup: I would keep it. I would like consistency across fletch for cxx 11/98, as much as possible. All that said, I feel like we should only support cxx11 compilers anyway, and likely make this an always on thing across fletch considering it's main purpose is to support kwiver. That is a larger discussion though and not for this PR.

Patches/OpenCV/3.1.0/common.cmake
dstoup: Thanks, perfect. It should show up when I run gitk, etc. I don't know what happens in the github UI though, but works for my purposes.

dstoup:
Looks good to me. I will cycle the dashboards again then merge. I resolved the conflicts with release-notes myself, using the web UI so I think that is the path forward for now.

dstoup:
Jenkins test this please

PR #175, opened 2017-08-04
Versioned VTK

Erotemic:
This PR exposes a variable `VTK_SELECT_VERSION` to select a version of VTK to build and updates the patches of VTK to use the versioning system proposed in #173. 

The default version of VTK is still 6.2, but the option now exists to set version 7.1.1 or version 8.0.


mleotta:
@RustyBlue, it looks like @Erotemic beat me too this, and his branch is more general than mine.  It supports changing the version.  You should be able to use this to test out VTK 8.0 for VIVIA.  As we discussed, the upgrade was fairly seamless for MAP-Tk.

mleotta:
@Erotemic I would drop the 7.1.1 unless there is some issue with 8.0 that requires us to fall back to an earlier release.  Let's keep 6.2 and 8.0 until we can confirm that we can drop 6.2.

Erotemic:
I have 7.1.1 in there because I'm not sure that OpenCV supports 8.0 yet. I'll drop it if I find out it does.

mleotta:
I didn't know OpenCV even used VTK :)

I'm not sure that ever use VTK via OpenCV, so that might not be a good driving factor.  We might be able to turn off VTK use in OpenCV if it's a problem, or patch OpenCV.

dstoup:
I didn't realize VTK was getting passed to OpenCV either. It is quite
likely the we can turn that off until someone clamours to have it.

On Aug 4, 2017 4:22 PM, "Matt Leotta" <notifications@github.com> wrote:

I didn't know OpenCV even used VTK :)

I'm not sure that ever use VTK via OpenCV, so that might not be a good
driving factor. We might be able to turn off VTK use in OpenCV if it's a
problem, or patch OpenCV.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Kitware/fletch/pull/175#issuecomment-320344634>, or mute
the thread

Erotemic:
PR #174 disables the OpenCV option to link to VTK by default. 

I believe the only thing this PR is doing right now is exposing 8.0 as an option (I removed 7.1.1 as an option), while still defaulting to the stable 6.2.

mleotta:
Assuming it doesn't cause any problems, I'm in favor of landing this branch and keeping the default VTK at 6.2.  This will give us some time to test VTK 8.0 with downstream applications that use VTK.  It will make it easier to do that testing by flipping a switch in Fletch to change VTK versions.  Once we are satisfied with VTK 8.0 we can make that the default in a separate PR.  And possibly remove support for VTK 6.2 as well.

mleotta:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
The same comment as before, the * Added experimental support for VTK version 8.0' release-note should go in the top section with package update info.

Ready to merge after that.

mleotta:
@dstoup it looks like the release notes updates are in the right section.

Erotemic:
Rebased on master. The release notes should be good now.

PR #177, opened 2017-08-04
Versioned Caffe

Erotemic:
I've been getting build failures when enabling cuda. Updating caffe to version 1.0 seems to fix this.

I made the Patch versioning changes to caffe. This one was a slightly tricky because caffe seemed to be being pulled from an internal build on `https://gitlab.kitware.com/kwiver/caffe/repository/archive.zip?ref=${Caffe_version}`. There was a version for windows and *nix. 

The 7f5cea3b2986a7d2c913b716eb524c27b6b2ba7b has was the *nix version and 527f97c0692f116ada7cb97eed8172ef7da05416 was the windows version. 

The same patch seemed to be applied to both, so I moved the existing patch into the 7f5cea3b... subfolder and then I duplicated the same code in the 527f97... subfolder. There might be a more elegant way (e.g symlinks) to do this when the same patch is applied to multiple versions, but I think it will ultimately cause less headaches to have each patch tied to a very specific version, even if that does duplicate some code (If only symlinks on windows were a thing). 

I'm also not sure if there is a reason why the BVLC github wasn't being targeted. If there is a reason why the Kitware gitlab Caffe is preferred let me know and I'll switch it back and see if I can patch the cuda issue.


mattdawkins:
Also D. Stoup just merged some branches which conflict with this

Erotemic:
I rebased and reverted to use the kitware hosted caffe by default.

Erotemic:
The most recent version now reverts to the internal kitware builds. It versions the patches by internal commit hash and it separates the windows the the non-windows patches. 

I also added a new patch to `ConfigGen.cmake` that fixes an issue when building with OpenCV 3.3.0.

dstoup:
LGTM. Are there any objections from Caffe users?

PR #182, opened 2017-08-11
Fix Caffe for Windows URL

hughed2:
The filename is not actually part of the download URL. As is, fletch breaks when it comes time to build caffe.


dstoup:
LGTM

PR #186, opened 2017-08-15
Versioned ffmpeg and added experimental support for 3.3.3

Erotemic:
This is part of #176 but I forgot to submit it as its own PR. 

When upgrading to OpenCV 3.3.0, I found that I needed to upgrade ffmepg. 

On *nix, by default the stable version 2.6.2 is still used, but the option for upgrading to 3.3.0 is now there. 

On windows nothing has changed. I'm not sure exactly how to deal with the windows case. It seems to pull down prebuilt binaries from the internal kitware repos which are built against git hash c089e72. This is not associated with any particular exact version (it seems to be just before 2.6). If updating to a new version on windows is as simple as uploading a new prebuilt version to `https://data.kitware.com`, I could do that, but I thought I'd ask before I did anything on something that appears to be very special-case. 

I versioned the patches, as I did with the other packages, but I made a special win32 folder for the patches to the prebuild windows binaries because the patch was windows specific. The alternative would be to make a `win32-c089e72` or `win32/c089e72` patch directory.
 
I also added some comments and documentation explaining the patches (because I had to figure out what was going on).


CMake/External_FFmpeg.cmake
```diff
@@ -47,6 +52,46 @@ else ()
   set(_FFmpeg_yasm --yasmexe=${fletch_YASM})
   list(APPEND ffmpeg_DEPENDS yasm)
 
+  # Should we try to point ffmpeg at zlib if we are building it?
```
dstoup: What happens if we don't have a system version?

Erotemic: The issue is that ffmpeg always seems to link against the system zlib even if we are building our own. I'm not sure if this is a problem or not. 

I'm not sure what happens if no system version is installed. The reason that I mentioned this is that when I build fletch with ffmpeg and zlib, and then I `ldd` the built libraries, I see that the ffmpeg libs point to the system version `libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1`, instead of `libz.so`, which is in the same build directory. 

I think this behavior is fine if we aren't building zlib, but should FFmpeg be pointing to the system version of the library if we are building it?

dstoup: I think we will have no choice but to hope for the best here. In looking at configure --help, it seems FFmpeg is one of those wonderful packages who don't let you pass a path to packages when you --enable-pkg, it insists on autodetect.

I think we just need to deal with it if it becomes an issue. Most *nix based systems should have what they need. The only thing we might be able to do it, find_package(ZLib .. ) and enable/disable based on that. I think what we have now is ok until it isn't.

PR #190, opened 2017-08-16
Fix OpenCV 3.1 linkage on MacOS

mleotta:
This patch is important for properly building MAP-Tk installers on MacOS.  The same fix was applied in Fletch to OpenCV 2.4 and upstream in OpenCV, but it did not land until OpenCV 3.2, so the fix is needed for 3.1.  Details here:

https://github.com/opencv/opencv/pull/6006

I know this will conflict with some OpenCV versioning changes on master, but it is important to get this change on the release branch.  We can resolve conflicts and put this patch only in the 3.1 patches in the new versioned approach.  3.2 and up should not need it.


mleotta:
@Erotemic just a heads up that there is another OpenCV 3.1 Patch on the way that will need conflict resolution with #174.  This issue is breaking things in the last release so I'm targeting the `release` branch, but this patch can eventually move in with the 3.1 patches in the new organization.  The patch is not needed with OpenCV 3.2 or above.

dstoup:
Looks good to me. I do remember the OpenCV 2 patch of the same thing.  I will merge as soon as the dashboards cycle. Given this change, I doubt the conflicts will be too serious. I might be able to resolve them myself.

PR #191, opened 2017-08-16
Don't create extra symlinks to libraries. The soversion property is enough

dstoup:
The patch fixes one of the two issues mentioned in issue #178.

There is a macro getting called which makes symlinks. When called on the libraries, it creates too many levels of symlink. It should not be needed anyway since we set_target_properties for the soversion, so the sylinks exists anyway.

The question I have is, should this PR go to master or release? Since it's a bugfix, I am inclined to target release. We can discuss and I will update release-notes once I know which one to edit. Also, I did take the branch from master, so if release is desired, I will need to rebase or cherry-pick.


dstoup:
Closing and reopening against release

PR #195, opened 2017-08-22
Don't create extra symlinks to libraries. The soversion property is g…

dstoup:
…ood.


dstoup:
Jenkins test this please

PR #206, opened 2017-08-29
New Darknet archive to fix a case issue in a file name

aaron-bray:
We are using 1 archive to server all platforms, I switched to the windows for all (as it has 3rd party pthread libs) and a file name was not cased properly and linux did not like that

### DIFF
--- a/CMake/fletch-tarballs.cmake
+++ b/CMake/fletch-tarballs.cmake
@@ -389,8 +389,8 @@ list(APPEND fletch_external_sources Caffe)
 
 # Darknet
 # The Darket package used is a fork maintained by kitware that uses CMake and supports building/running on windows
-set(Darknet_url "https://data.kitware.com/api/v1/file/598866018d777f7d33e9c0f3/download/darknet-win32-70dfcb43.zip")
-set(Darknet_md5 "eea312b07685eef9c6f26cf63346fe41")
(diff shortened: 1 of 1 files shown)


dstoup:
Do we need a new archive to fix 1 file? Would a patch be sufficient or is it more complicated than it's worth?

aaron-bray:
A patch would work, but I would have to setup patch logic in the External_Darknet.config and hook it up to get everything working. 
The only issue is changing darknet.config.in to Darknet.config.in
Putting in the infrastructure to patch darknet does not seem worth it as we control the repo we are pulling from anyway, there really should be no need to patch darket...

Seems easier just to replace the archive

dstoup:
My concern would be, if we do code updates relying on a second archive adds maintenance issues, one more thing that needs to get done.

If we control the repo anyway, can't we just fix the issue in the code so we are looking for the correct .in file?

aaron-bray:
it should be upper case 'D' as find_package is looking for Darknet and not darknet
fletch is correct in refering to 'D'arknet, I would rather not change code in fletch to be inconsistent in casing in different places. That may cause confusion

I just missed the uppercasing of 'd'arknet.config.in on windows, because windows is case insensitive

This single archive serves all platforms ( I am going to delete the other 2 archives)

dstoup:
Sounds good, thanks!

dstoup:
Jenkins test this please

PR #208, opened 2017-08-30
Dev/fix gflags header

hughed2:
This fixes an error caused by GFlags switching namespace.


dstoup:
I haven't tested but, is this an issue in 3.3 too?

hughed2:
No, this had already been fixed in OpenCV_contrib by 3.3.

PR #210, opened 2017-09-07
Patch opencv module sfm

dstoup:


dstoup:
@mleotta .. this should fix the issue you reported.

PR #211, opened 2017-09-10
Revert CUDA enable flag name

mattdawkins:
fletch_BUILD_WITH_CUDA was changed to fletch_ENABLE_CUDA, but it wasn't changed for all projects (e.g. CAFFE) so there's a bug in master since some things were using 1 flag but not the other.

I prefer BUILD_WITH_CUDA as opposed to ENABLE_CUDA because unlike all of the other packages, CUDA is not being built by fletch, just linked against, and the ENABLE_* would make it seems like it was building it like everything else.

### DIFF
--- a/CMake/External_Darknet.cmake
+++ b/CMake/External_Darknet.cmake
@@ -18,13 +18,13 @@ else()
   set(DARKNET_OPENCV_ARGS -DUSE_OPENCV:BOOL=OFF)
 endif()
 
-if(fletch_ENABLE_CUDA)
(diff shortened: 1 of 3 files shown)


mattdawkins:
@aaron-bray thoughts?

abray3:
Yeah, that makes sense to me

dstoup:
Jenkins test this please

dstoup:
Makes sense to me too, thanks for the patch!

PR #216, opened 2017-09-15
Account for change in MSVC_VER for patched vs2017

dstoup:


dstoup:
Linux test is not relevant in this case. Merging since it fixes a bug.

PR #217, opened 2017-09-15
Fix OpenCV Python with Win32 Debug builds

mleotta:
This fix works and is probably good enough for us.  A slightly better approach would be to use CMake to detect if the Python Debug libraries are found and only employ this hack if they are not.  I may try that approach in the future and push it upstream to OpenCV.

### DIFF
--- a/Doc/release-notes/release.txt
+++ b/Doc/release-notes/release.txt
@@ -32,6 +32,10 @@ OpenCV 3.1
    in Fletch and was accepted upstream in OpenCV 3.2.  Details here:
    https://github.com/opencv/opencv/pull/6006
 
+ * Patched OpenCV 3.1 to fix Debug builds on Windows when Python is enabled.
+   The issue was failure to link to python27_d.lib, which is not distributed
+   with Python installers.
+
 PNG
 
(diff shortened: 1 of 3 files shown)


dstoup:
Do we need to patch this for other versions of OpenCV?

mleotta:
Possibly, but I don't have time to test that.

mleotta:
A side note on this PR:  Once this lands in `release`, we need to be careful about merging `release` into `master`.  The new `cv2.cpp` added here in `Patches\OpenCV3\` needs to move to `Patches\OpenCV\3.1.0\` on `master`.  By default this will file will merge without conflict and not properly relocate causing `master` to break.  The file needs to be manually moved when merged into `master`.

dstoup:
Sounds good. I can try picking at the others as the problem arises. At least we have the template for the fix.

mleotta:
@dstoup This branch was to fix Windows builds and it works on all my Windows configurations.  The Jenkins failure appears to be due to that Qt issue you had after upgrading the compiler.  Has that been fixed?  Can we just merge this PR?  Do you want to re-test, and if so, do I need to rebase to get your fix for Qt?

dstoup:
The Qt PR merged. I don't think we will need a rebase. I would say, go
ahead and merge. If you can't, I will when I get to the hotel.

PR #220, opened 2017-09-21
Patched OpenCV version 3.1.0 to work with MSCV 1911 and darknet

Erotemic:
This patch was developed by @aaron-bray. I separated it into a standalone PR.


Erotemic:
Closed in favor of #221

PR #224, opened 2017-09-22
Provide an explicit download name for all packages without one.

dstoup:
For many packages, the default download package is just a version number which makes it difficult to know which downloaded file goes with which package.


dstoup:
Jenkins test this please

PR #225, opened 2017-09-25
Hide Protobuf /FFmpeg select version when disabled

Erotemic:
@dstoup with respect to your comments from #222, after I updated to the most recent master I noticed that the Protobuf_SELECT_VERSION and FFmpeg_SELECT_VERSION options were visible even if their corresponding enable flag was false. This change fixes that so they only become visible once they are enabled. (The reason I didn't notice this with FFmpeg before was because its select version is an advanced option).


dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

PR #230, opened 2017-09-27
Update fletch config with CUDA and CUDNN locations (if enabled)

aaron-bray:
Also added a new zip for Darknet to girder


dstoup:
Jenkins test this please

PR #231, opened 2017-09-28
logic to append all _SELECT_VERSION vars to the config

Erotemic:
This ensures that the selected version information for enabled packages is propagated to the fletchConfig.cmake file. 

I'm not sure if its more desirable to do it implicitly like this, or explicitly in the External_*.cmake packages. However, the "_SELECT_VERSION" variable is often used to set the value of a normal "<pkg>_version" variable, which is used in those scripts. Perhaps the two should be consolidated? If not, then I think the implicit version is correct, because we are assuming that all user options regarding version will have the "_SELECT_VERSION" suffix.


dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Looks reasonable. Waiting for dashboards to finish, then will merge.

PR #234, opened 2017-10-04
Cuda support

aaron-bray:
Use the cuda FindCUDA finds
Allow user to change it
Made sure kwiver uses the same CUDA if changed


dstoup:
Jenkins test this please

PR #239, opened 2017-10-06
Fix for Ceres and Caffe to share the fletch glog and gflags

aaron-bray:
Also a small fix to give Ceres a hint to find eigen

Note I converted tabs to spaces in the WindowsDownloadPrebuiltDependencies.cmake
The actual needed change was ~line 80
PATTERN "glog*" EXCLUDE
PATTERN "gflag*" EXCLUDE

I had to patch FindGlog because the find_package(glog) would find the glog-config.cmake but for whatever reason, it would NOT find the glog library. So I did this patch.
The find_package(gflags) in FindGflags.cmake uses gflags-config.cmake and finds the gflags lib, 
but glog.lib isnot found when find_package(glog) uses the glog-config.cmake

### DIFF
--- a/CMake/External_Caffe.cmake
+++ b/CMake/External_Caffe.cmake
@@ -47,8 +47,6 @@ endfunction()
 
 # Check for dependencies.
 if(NOT WIN32) # Win32 build takes care of most dependencies automatically
-  addCaffeDendency(GFlags "")
-  addCaffeDendency(GLog "")
   addCaffeDendency(LevelDB "")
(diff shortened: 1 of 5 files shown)


dstoup:
Jenkins test this please

aaron-bray:
If we merge this in, does kwiver automatically use this in it's nightly build?

hughed2:
Yes, once the fletch nightly runs and succeeds. The only time we would need to worry about nightly build updating is if there's a config change.

PR #243, opened 2017-10-10
Also trigger on caffe auto dep option

mattdawkins:


dstoup:
Jenkins test this please

PR #248, opened 2017-10-13
Caffe segnet

aaron-bray:
I added a zip for this repository to girder : https://github.com/TimoSaemann/caffe-segnet-cudnn5
(Note that maybe out of date due to : https://github.com/TimoSaemann/ENet)

Anyway
What cool in this PR, is the patch, while all the caffe cmake files, allows you to uniquely identify your caffe fork when you do an external_project_add, provide a fork name (ex. -DCAFFE_FORK_SUFFIX:STRING=Segnet)
And the cmake patch files will post fix this to all the targets, configurations, install directories, etc. that it creates.
For this Segnet fork of caffe, it will create/install to the following:
fletch/install/headers/caffe_segnet
libraries and example binaries named : caffe_segnet
the cmake configuration name is updated to Caffe_SegnetConfig.cmake

Of course this will not work if the CMake files are changed, but these cmake files glob up source and headers, so I do not think they are likely to change frequently, and between forks. But we can patch if they do. 

I think we can use these cmake files in our kwiver/caffe fork and name it something like caffe_kw, so if somebody wants to use a different caffe fork, they don't have to name change theirs, theirs would make a regular default named 'caffe' binary, library, header, configuration in fletch.

Food for thought.

### DIFF
--- a/CMake/External_Caffe_Segnet.cmake
+++ b/CMake/External_Caffe_Segnet.cmake
@@ -1,0 +1,350 @@
+set(allOk True)
+set(errorMessage)
+
+option(AUTO_ENABLE_CAFFE_SEGNET_DEPENDENCY "Automatically turn on all caffe segnet dependencies if caffe segnet is enabled" OFF)
+if(fletch_ENABLE_Caffe_Segnet AND AUTO_ENABLE_CAFFE_SEGNET_DEPENDENCY)
+  #Snappy is needed by LevelDB and ZLib is needed by HDF5
+  if(WIN32)
+    set(dependency Boost ZLib OpenCV HDF5)
+  else()
+    set(dependency Boost GFlags GLog ZLib HDF5 Snappy LevelDB LMDB OpenCV Protobuf)
+  endif()
+
+  if(NOT APPLE AND NOT WIN32)
+    list(APPEND dependency OpenBLAS)
+  endif()
+
+  set(OneWasOff FALSE)
+
+  foreach (_var IN LISTS dependency)
+    get_property(currentHelpString CACHE "fletch_ENABLE_${_var}" PROPERTY HELPSTRING)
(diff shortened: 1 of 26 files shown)


dstoup:
Jenkins test this please

aaron-bray:
Windows is now safe guarded.
Caffe-Segnet does not build on windows yet.
We will probably need to make a fork in gitlab/kwiver and do the same changes we did on gitlab/kwiver/caffe to get it to build on windows

dstoup:
Jenkins test this please

CMake/External_Caffe_Segnet.cmake
```diff
@@ -0,0 +1,356 @@
...
+    PACKAGE Caffe_Segnet
+    PACKAGE_DEPENDENCY ${depend}
+    PACKAGE_DEPENDENCY_ALIAS ${depend}
+    )
+endfunction()
+
+# Check for dependencies.
+if(NOT WIN32) # Win32 build takes care of most dependencies automatically
+  addCaffeSegnetDendency(GFlags "")
+  addCaffeSegnetDendency(GLog "")
```
dstoup: Both GLog and GTest are available on Window now and included in Caffe Windows.

aaron-bray:
Here is the diff with the original files
Good point on putting the original files in, then applying the changes
(take off the .log extension if you have a nice diff viewer)

[Caffe-Segnet.diff.log](https://github.com/Kitware/fletch/files/1388033/Caffe-Segnet.diff.log)

CMake/External_Caffe_Segnet.cmake
aaron-bray: Ok, but FYI
We still need to go through the actual code base and make it windows compatible much like the kwiver gitlab caffe version we have.

dstoup:
Jenkins test this please

PR #254, opened 2017-10-24
Ceres requires Suitesparse with requires lapack (on ubuntu 16)

aaron-bray:
I needed to do this to get fletch build on a clean linux


dstoup:
This requirement is everywhere. Thanks for the patch!

PR #261, opened 2017-11-02
Geotiff force fletch proj4

dstoup:
Make the add_package_dependency of PROJ4 optional.

The system version seems to have issues on several linux distributions. Libraries are missing symbols which makes it  impossible for geotiff to link.
 
We are making the add_package_dependency optional to avoid the confusing error message about using the system version. We can safely make this call optional since we have the fatal error before it which forces enabling proj4.


dstoup:
@Erotemic can you test this branch and make sure it fixes your issue #259 ?

PR #273, opened 2017-11-27
Patch ffmpeg regex escape

dstoup:


dstoup:
Linux failure is unrelated to this branch.

PR #278, opened 2017-12-06
Yasm required python

dstoup:
This branch was created because yasm silently requires python to build. It does the following:

1. Moves the find python and version logic to a common macro
2. Ensures that the PYTHON_EXECUTABLE is found allowing packages which require yasm
3. If PYTHON_EXECUTABLE is not found, log message and attempt to find it with new macro.
4. Pass the PYTHON_EXECUTABLE to yasm's build.

It is important to note, this branch does not flip fletch_BUILD_WITH_PYTHON, and restores the value of PYTHON_EXECUTABLE if it was modified. That prevents all packages building their python components when it wasn't enabled.


CMake/External_libjpeg-turbo.cmake
```diff
@@ -1,5 +1,17 @@
 # libjpeg-turbo External project
 
+if (NOT PYTHON_EXECUTABLE)
+  message(STATUS "Building libjpeg-turbo requires yasm which requires python to build. Searching for PYTHON_EXECUTABLE")
+  include(CommonFindMacros)
+  fletch_FIND_PYTHON()
```
mleotta: Instead of finding PYTHON here we need to ask the user to explicitly set `yasm_PYTHON_EXECUTABLE`  we also need an else case here that sets `yasm_PYTHON_EXECUTABLE` to `PYTHON_EXCUTABLE` in the case that is is found.

PR #309, opened 2018-01-19
Leveldb cmake

dstoup:
Not the best, but it does work. Test the basic functionality and I will get it testing on dashboards. The file does try to support Windows.


dstoup:
@aaron-bray

PR #312, opened 2018-01-26
Patch levelDB for older gcc MAX_SIZE def

dstoup:


dstoup:
Jenkins test this please

PR #324, opened 2018-02-16
Don't processess any FFmpeg config unles it's enabled

dstoup:


dstoup:
@billhoffman see if this change fixes your issue.

billhoffman:
It works for me.

PR #325, opened 2018-02-16
Add pybind11 patches

KyleFromKitware:
This adds a set of patches to pybind11 that fix pybind/pybind11#1276 (the changes I've made can be seen in pybind/pybind11#1286)

### DIFF
--- a/CMake/External_pybind11.cmake
+++ b/CMake/External_pybind11.cmake
@@ -13,6 +13,11 @@ ExternalProject_Add(pybind11
   PREFIX ${fletch_BUILD_PREFIX}
   DOWNLOAD_DIR ${fletch_DOWNLOAD_DIR}
   INSTALL_DIR ${fletch_BUILD_INSTALL_PREFIX}
+  PATCH_COMMAND ${CMAKE_COMMAND}
+    -Dpybind11_source:STRING=${fletch_BUILD_PREFIX}/src/pybind11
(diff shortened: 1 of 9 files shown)


dstoup:
A comment about patch branches in general. I prefer 2 commits. One which pushes the original files and a second which applies the patch. That will allow reviewers to see what's actually being changed.

KyleFromKitware:
Like this?

dstoup:
Yes, thank you :)

dstoup:
Jenkins test this please

KyleFromKitware:
The builds are failing:

```
CMake Error at CMake/External_Tensorflow.cmake:4 (message):
  Error: A build with Python is required for building Tensorflow
```

Is Jenkins not configured for Python? Does this usually happen?

dstoup:
FYI, this is more for Dawkins than anyone, but if there are useful changes being made, I would prefer they go into master, not some frankenbranch which will not merge. 

Are these changes useful for pybind11 in general?

hughed2:
Kyle, do you want to push that into viame/master or plain master? Jenkins is configured for master, if viame/master needs anything different I wouldn't expect it to work.

dstoup:
Jenkins is not configured with python and any packages which are added need to assume it can be turned on or off.

dstoup:
@hughed2 this branch cannot go into master, it was taken from viame/master.

PR #339, opened 2018-03-01
Dev/fix opencv gotoblas

Erotemic:
OpenCV 3.4 has an option `OpenBLAS_LIB`. I'm not sure if it exists in previous versions. ~~I need to check that in order to correctly set the version if statement~~ ( Turns out OpenBLAS support was added in `3.2`, so I happened to guess right. ). The first function of this PR is to pass down that info when it is needed. 

The second function is to fix an issue I get:
```
    [ 97%] Linking CXX executable ../../bin/opencv_traincascade
    //usr/lib/liblapack.so.3: undefined reference to `gotoblas`
    collect2: error: ld returned 1 exit status
    apps/traincascade/CMakeFiles/opencv_traincascade.dir/build.make:392: recipe for target 'bin/opencv_traincascade' failed
```

A workaround for this seems to be to set `WITH_LAPACK=False`. It would be better if I understood the issue more, but I don't.

### DIFF
--- a/CMake/External_OpenBLAS.cmake
+++ b/CMake/External_OpenBLAS.cmake
@@ -44,11 +44,20 @@ endif()
 
 fletch_external_project_force_install(PACKAGE OpenBLAS)
 
+
 set(OpenBLAS_ROOT ${fletch_BUILD_INSTALL_PREFIX} CACHE STRING "")
+
+# Find the OpenBLAS library name that will be passed to OpenCV / Caffe / etc..
+# Note: the lib does not exist yet, but we can be pretty sure what it will be
+get_system_library_name(openblas openblas_libname)
+set(OpenBLAS_LIB "${OpenBLAS_ROOT}/lib/${openblas_libname}")
+set(OpenBLAS_INCLUDE_DIR "${OpenBLAS_ROOT}/include")
 
 file(APPEND ${fletch_CONFIG_INPUT} "
 ########################################
 # OpenBLAS
 ########################################
 set(OpenBLAS_ROOT    \${fletch_ROOT})
+set(OpenBLAS_LIB     \${OpenBLAS_LIB})
+set(OpenBLAS_INCLUDE_DIR \${OpenBLAS_INCLUDE_DIR})
 ")
--- a/CMake/External_OpenCV.cmake
+++ b/CMake/External_OpenCV.cmake
@@ -86,11 +86,15 @@ if(fletch_ENABLE_OpenBLAS)
   if (NOT OpenCV_version VERSION_LESS 3.2.0)
     message(STATUS "OpenCV depending on fletch OpenBLAS")
     set(_OpenCV_ENABLE_OPENBLAS_DEFAULT TRUE)
-    #list(APPEND OpenCV_DEPENDS OpenBLAS)  # complains that this target doesnt exist. Does it?
+    list(APPEND OpenCV_DEPENDS OpenBLAS)
     list(APPEND OpenCV_EXTRA_BUILD_FLAGS
-      -DWITH_LAPACK:BOOL=FALSE  # workaround for undefined referenec to `gotoblas`
+      # FIXME: If LAPACK is on, opencv_traincascade fails to link.
+      # The error is: liblapack.so - undefined reference to `gotoblas`.
+      # note: the problem may be multiple cblas.h
+      # A temporary workaround is to turn LAPACK off.
+      -DWITH_LAPACK:BOOL=FALSE
       -DOpenBLAS_INCLUDE_DIR:PATH="${OpenBLAS_ROOT}/include"
-      -DOpenBLAS_LIB:PATH="${OpenBLAS_ROOT}/lib/${openblas_libname}"
+      #-DOpenBLAS_LIB:PATH="${OpenBLAS_ROOT}/lib/${openblas_libname}"
       )
   endif()
 endif()
--- a/CMake/External_OpenCV.cmake
+++ b/CMake/External_OpenCV.cmake
@@ -83,8 +83,7 @@ endif()
 
 
 if(fletch_ENABLE_OpenBLAS)
-  if (OpenCV_version VERSION_LESS 3.2.0)
-  else()
+  if (NOT OpenCV_version VERSION_LESS 3.2.0)
     message(STATUS "OpenCV depending on fletch OpenBLAS")
     set(_OpenCV_ENABLE_OPENBLAS_DEFAULT TRUE)
     #list(APPEND OpenCV_DEPENDS OpenBLAS)  # complains that this target doesnt exist. Does it?
--- a/CMake/External_OpenCV.cmake
+++ b/CMake/External_OpenCV.cmake
@@ -82,6 +82,22 @@ else()
 endif()
 
 
+if(fletch_ENABLE_OpenBLAS)
+  if (OpenCV_version VERSION_LESS 3.2.0)
+  else()
+    message(STATUS "OpenCV depending on fletch OpenBLAS")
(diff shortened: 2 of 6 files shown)


CMake/External_OpenCV.cmake
```diff
@@ -82,6 +82,21 @@ else()
...
+if(fletch_ENABLE_OpenBLAS)
+  if (OpenCV_version VERSION_LESS 3.2.0)
+  else()
+    message(STATUS "OpenCV depending on fletch OpenBLAS")
+    set(_OpenCV_ENABLE_OPENBLAS_DEFAULT TRUE)
+    #list(APPEND OpenCV_DEPENDS OpenBLAS)  # complains that this target doesnt exist. Does it?
+    list(APPEND OpenCV_EXTRA_BUILD_FLAGS
+      -DWITH_LAPACK:BOOL=FALSE  # workaround for undefined referenec to `gotoblas`
+      -DOpenBLAS_INCLUDE_DIR:PATH="${OpenBLAS_ROOT}/include"
+      -DOpenBLAS_LIB:PATH="${OpenBLAS_ROOT}/lib/${openblas_libname}"
```
dstoup: I don't see openblas_libname defined in this file. I suppose if it's working, we might be getting lucky and picking it up from the Caffe configuration. Better to have it in here in case someone only builds openBLAS and OpenCV.

CMake/External_OpenCV.cmake
```diff
@@ -82,6 +82,21 @@ else()
 endif()
 
 
+if(fletch_ENABLE_OpenBLAS)
+  if (OpenCV_version VERSION_LESS 3.2.0)
```
dstoup: Why not just use 
`if (NOT OpenCV_version VERSION_LESS 3.2.0)`
and skip the empty if statement?

Erotemic: Because my knowledge of CMake Scripting syntax is a bit lacking, but now it is less so.

Erotemic:
I was able to get OpenCV to depend on OpenBLAS by moving the OpenBLAS tarball block to appear before OpenCV.

CMake/External_OpenBLAS.cmake
```diff
@@ -44,11 +44,20 @@ endif()
...
+set(OpenBLAS_LIB "${OpenBLAS_ROOT}/lib/${openblas_libname}")
+set(OpenBLAS_INCLUDE_DIR "${OpenBLAS_ROOT}/include")
+
 file(APPEND ${fletch_CONFIG_INPUT} "
 ########################################
 # OpenBLAS
 ########################################
 set(OpenBLAS_ROOT    \${fletch_ROOT})
+set(OpenBLAS_LIB     \${OpenBLAS_LIB})
+set(OpenBLAS_INCLUDE_DIR \${OpenBLAS_INCLUDE_DIR})
```
Erotemic: @dstoup I'm unsure if these should go here.

CMake/External_OpenCV.cmake
```diff
@@ -82,6 +82,24 @@ else()
...
+  if (NOT OpenCV_version VERSION_LESS 3.2.0)
+    message(STATUS "OpenCV depending on fletch OpenBLAS")
+    set(_OpenCV_ENABLE_OPENBLAS_DEFAULT TRUE)
+    list(APPEND OpenCV_DEPENDS OpenBLAS)
+    list(APPEND OpenCV_EXTRA_BUILD_FLAGS
+      # FIXME: If LAPACK is on, opencv_traincascade fails to link.
+      # The error is: liblapack.so - undefined reference to `gotoblas`.
+      # note: the problem may be multiple cblas.h
+      # A temporary workaround is to turn LAPACK off.
+      -DWITH_LAPACK:BOOL=FALSE
```
Erotemic: If anyone has a guess as to what is causing this issue I'd love to know. Otherwise this workaround is probably fine for the time-being.

dstoup:
Jenkins test this please

CMake/External_OpenBLAS.cmake
dstoup: I think this is the right location for these variables.

CMake/External_OpenCV.cmake
dstoup: What is lapack providing to OpenCV? Just wondering what we're losing by disabling. One option we'd considered was having fletch provide lapack, which I think is more a larger suite than a single tool, but I can't recall at the moment.

CMake/fletch-tarballs.cmake
```diff
@@ -154,6 +154,23 @@ set(Eigen_md5 "6a578dba42d1c578d531ab5b6fa3f741")
 set(Eigen_dlname "eigen-${Eigen_version}.tar.gz")
 list(APPEND fletch_external_sources Eigen)
 
+#OpenBLAS
+if(NOT WIN32)
+  #set(OpenBLAS_version "0.2.15")
```
dstoup: dead code, remove

CMake/fletch-tarballs.cmake
```diff
@@ -154,6 +154,23 @@ set(Eigen_md5 "6a578dba42d1c578d531ab5b6fa3f741")
...
+  set(OpenBLAS_version "0.2.20")
+  set(OpenBLAS_url "https://github.com/xianyi/OpenBLAS/archive/v${OpenBLAS_version}.tar.gz")
+
+  if (OpenBLAS_version VERSION_EQUAL 0.2.20)
+    set(OpenBLAS_md5 "48637eb29f5b492b91459175dcc574b1")
+  elseif (OpenBLAS_version VERSION_EQUAL 0.2.15)
+    set(OpenBLAS_md5 "b1190f3d3471685f17cfd1ec1d252ac9")
```
dstoup: Do we really need both packages? OpenBLAS is more of an incidental package, not used directly by out algorithms.

dstoup:
FYI, Windows didn't fail, I just stopped it while we examine review comments. We have a stack of Windows builds to get through.

CMake/fletch-tarballs.cmake
Erotemic: probably not.

CMake/External_OpenCV.cmake
Erotemic: I don't know, but I'd like to know the answer as well. I think OpenBLAS contains the LAPACK (Linear Algebra PACKage) functions. I'm not sure if this completely disables BLAS or just the LAPACK functions.  I'm really not sure what the exact structure of the BLAS+LAPACK libraries are. Does LAPACK exist as a standalone library or does it always exist as a part of a BLAS (e.g. OpenBLAS, ATLAS, MKL) package?

CMake/fletch-tarballs.cmake
Erotemic: done

CMake/External_OpenBLAS.cmake
Erotemic: Other packages don't seem to output these variables in the `fletch_CONFIG_INPUT`. Is it the case that the other packages should be doing this? Or are packages free to export however much they want?

dstoup: I was having this conversation with someone recently. I think it makes like easier if all of the knowledge about a package is generated in its configuration. That eliminates the duplication of get_library_name type stuff.

CMake/External_OpenCV.cmake
dstoup: I am not able to reproduce the issue with gotoblas. Is there another PR that deals with the OpenBLAS addition? I would like to work on that part first and see if we can figure out what's going on with the error.

dstoup:
Jenkins test this please

PR #343, opened 2018-03-01
Disable export to package registry

### LINKED ISSUE #341: Fletch Packages Found by External Projects --- Unintended Interactions with The CMake Package Registry?
So, I've just spent a better part of an evening trying to figure out why a external build of caffe2 keeps finding packages in one of my random fletch build directories. This particular fletch build was not in my CMAKE_PREFIX_PATH, PATH, CPATH, or any other environment variable I checked. However, whenever I configured caffe2 it kept returning directories within this specific fletch build when they called find_packages on eigen3, gflags, and glog. 

After a lot of debugging I finally figured out it was due to a feature of CMake called the CMake Package Registry (which was new to me). Apparently the Caffe  Caffe_Segnet  gflags  glog and  yaml-cpp packages are writing information there. (On Linux the registry is implemented as files in the `~/.cmake/packages` directory, so deleting that directory solves the problem temporarily).
 
For someone like myself who has a good deal of CMake experience, but was not awar e of this quirk, this behavior is highly non-intuitive. 

Is it something within Fletch or these packages that is causing this package registry to be written? Should we turn it off? If so would it be as simple as setting `CMAKE_EXPORT_NO_PACKAGE_REGISTRY=True` in the fletch CMakeLists.txt?

Erotemic:
Attempts to fix #341

I'm not sure if this is sufficient or we need to pass down this setting to all subprojects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,10 @@ if (fletch_BUILD_CXX11)
   set(fletch_CXX_STANDARD_VERSION "11")
 endif()
 
+
+# Disable exporting to the cmake package registry (see GH #341)
+set(CMAKE_EXPORT_NO_PACKAGE_REGISTRY TRUE)
+
 #
 # Do we want to build in Python support where available
 #


dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

PR #344, opened 2018-03-01
Add package specific variables to External_*.cmake

Erotemic:
Applies the pattern used in #339 to define variables relevant to a package in their `External_*.cmake`. This PR just defines the variables (so I can use them in my caffe2 branch), but does not simplify the corresponding places in the Caffe and Caffe_Segnet build scripts. This does cause temporary duplication of code, but it also makes it easier to review.


dstoup:
These changes look good. There should be a corresponding change to remove the duplicate calls from the External_*Caffe* cmake files.

PR #372, opened 2018-05-11
Update the default gdal version to 2.2.2

dstoup:


dstoup:
Covered in a more recent branch.

PR #375, opened 2018-05-21
On Windows TinyXML is built without STL, breaking downstream

RustyBlue:
The `readme.txt` for the TinyXML source indicates that compiling with STL support on Windows, which is assumed by some projects using TinyXML, requires adding `#define TIXML_USE_STL` "as the first line in tinyxml.h". This change to the patch copies the `tinyxml.h` with the appropriate define.


dstoup:
I'm confused as to why it is configuring without STL. I put a nontrivial
amount of effort into making the try/compile work in debug and release on
Windows so that it actually detected STL stand configured accordingly.

Let's chat tomorrow so we can figure out why this is happening.

RustyBlue:
Issue is (was?) downstream.

PR #387, opened 2018-06-15
Fixed the setting of `qtExtensions_DIR`.

chetnieter:
Fixed `qtExtensions_DIR` so it points to the correct directory that actually contains `qtExtensionsConfig.cmake`.


dstoup:
Jenkins test this please

PR #395, opened 2018-06-27
Qt5 default

dstoup:
This branch also includes the VTK to Qt5 branch which is currently out for review. The main purpose in this MR for the moment is that it is a way to test that MR with Qt5 enabled.


mleotta:
FWIW, my suggestion is to make the next release of Fletch fully support the option to build Qt4 or Qt5.  However, we should leave Qt4 as the default in the upcoming release.  After making the release, we should probably make Qt5 the default in `master`, but continue to support Qt4 as an option for a while.

I realize this branch is more about triggering dashboard tests for now, but this seemed like the right place to make this comment.

dstoup:
Yes, you're correct. There was no intention of merging this branch for the
release, just looking for a way to fully test the VTK branch.

mleotta:
You may want to merge in or rebase this on https://github.com/Kitware/fletch/pull/398

dstoup:
Jenkins test this please

PR #396, opened 2018-06-29
Python command fail reset numpy

dstoup:
We already reset the numpy find variables if the version test fails, but we also need to reset if the entire command fails. If we don't, the user will need to remove the build directory and start again if numpy is installed after. The current behaviour is very confusing. I got the error, installed numpy and still got the error. Resetting the variables lets the user fix the existing configuration rather than delete.


aaron-bray:
It makes sense to unset it.
And you have to unset after the execute_process to clear it from anything being set if it did not find it.

I will be working through some python/numpy code later this month and will keep this in mind

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

PR #408, opened 2018-07-11
Add Extra Flag for OpenCV Using TIFF

mattdawkins:


CMake/External_OpenCV.cmake
```diff
@@ -107,31 +107,40 @@ else()
...
 
-  list(APPEND OpenCV_EXTRA_BUILD_FLAGS
-    -DBUILD_TIFF:BOOL=OFF
-    -DTIFF_INCLUDE_DIR:PATH=${TIFF_INCLUDE_DIR}
-    -DTIFF_LIBRARY_DEBUG:FILEPATH=${TIFF_LIBRARY_DEBUG}
-    -DTIFF_LIBRARY_RELEASE:FILEPATH=${TIFF_LIBRARY_RELEASE}
-    )
+# Allow OpenCV's tiff support to be turned off
+option(fletch_ENABLE_OpenCV_TIFF "Build OpenCV with tiff image support" TRUE )
+mark_as_advanced(fletch_ENABLE_OpenCV_TIFF)
```
dstoup: I'm not really in favor of these kinds of arguments. It opens a an endless permutations for every package that can deal with every image format. We've always followed the practice that if you enable something in fletch support for it is added. Is there a good reason to force disable? If so, there needs to be a better approach (better engineering) in place that doesn't involve adding these flags for every package. This type of thing requires a larger discussion.

mattdawkins: People were reporting build issues in VIAME because of this on a few linux platforms, leaving it turned off resolved said issues, because tiff support in opencv is un-needed in VIAME

mattdawkins: The errors themselves were primarily linking errors on Ubuntu / Fedora, didn't have time to deal with at the time and still don't, it was better for us to just leave it disabled since it's un-needed

dstoup: Link errors from viame/master or master? There are a lot of fixes in master which relate to this problem.

dstoup: If you can point me to an instance of a build failure, I will take a look at fixing this the right way, if possible.

mattdawkins: viame/master on fletch is close to master, so both. Reproducing is low priority right now, won't happen for a bit

PR #409, opened 2018-07-11
Geotiff is not a hard dependency of VXL

mattdawkins:


CMake/External_VXL.cmake
```diff
@@ -20,11 +20,13 @@ add_package_dependency(
   )
 
 # libgeotiff
-add_package_dependency(
-  PACKAGE VXL
-  PACKAGE_DEPENDENCY libgeotiff
-  PACKAGE_DEPENDENCY_ALIAS GEOTIFF
-  )
```
dstoup: I think you can enable the same thing by using the OPTIONAL option in add_package_dependency, right?

mattdawkins: I'm not sure on the exact particulars of the script, ideally though we want to force geotiffs use if it is turned on at least, does the script do that?

dstoup:
Yes. OPTIONAL should enable it if fletch builds it or it's found on the system but not error out if geotiff wasn't found.

dstoup:
Since this branch is a good idea but you haven't fixed it, I am closing. There is a new branch which does this correctly. Have a look at #469

PR #411, opened 2018-07-11
Add Extra Symlink for Boost Python

mattdawkins:
This change should become deprecated in the future as boost python is removed, but in the current state it's still useful for a few projects when building with Python3 as opposed to Python2 (with 3 it get's generated as boost_python3.so vs just boost_python.so which messes up linking in Caffe and other packages).


dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

PR #417, opened 2018-07-18
Update openjpeg2 version to 2.3.0

Purg:
This is primarily due to an issue with GDAL where its if-defs are
improperly written such that if the host system has openjpeg version 2.2
or greater, GDAL sees those header files first and activates features
not available in 2.1.


Purg:
PTAL @dstoup

dstoup:
Jenkins test this please

PR #419, opened 2018-07-19
Improve handling of Boost_EXTRA_LIBS

zackgalbreath:
This list of extra Boost libraries needs to be passed to the configure
script where BCP is run.  This commit also makes this list a cache variable
so it's easier for users to specify what extra Boost libraries they need.


dstoup:
Jenkins test this please

dstoup:
Thanks @zackgalbreath !

PR #424, opened 2018-08-01
Disable CMake Registry passes to all CMake packages

dstoup:


mleotta:
I have not tested it, but this looks like the right way to fix it.

PR #426, opened 2018-08-07
Pass the fletch proj4 to GDAL

dstoup:


dstoup:
Power outage caused the Windows build to fail.

PR #428, opened 2018-08-08
added zeromq

Cookt2:


Cookt2:
Jenkins test this please

Cookt2:
Jenkins test this please

CMake/External_cppzmq.cmake
```diff
@@ -0,0 +1,32 @@
...
+# dependency to "install" the cppzmq header. If ZeroMQ is not enabled, a
+# warning is shown.
+
+if(NOT fletch_ENABLE_ZeroMQ)
+  message(WARNING "cppzmq module enabled without ZeroMQ! Only the header will be installed!")
+endif()
+
+ExternalProject_Add(cppzmq
+  PREFIX ${fletch_BUILD_PREFIX}
+  GIT_REPOSITORY ${cppzmq_git_repo}
```
dstoup: We don't want to use git. Download a tarfile.

CMake/External_cppzmq.cmake
```diff
@@ -0,0 +1,32 @@
...
+    -P ${fletch_SOURCE_DIR}/Patches/cppzmq/Patch.cmake
+  CMAKE_GENERATOR ${gen}
+  CMAKE_ARGS
+    -DCMAKE_INSTALL_PREFIX:PATH=${fletch_BUILD_INSTALL_PREFIX}
+    -DBUILD_SHARED_LIBS:BOOL=${BUILD_SHARED_LIBS}
+  BUILD_IN_SOURCE 1
+)
+
+# CMake build configuration additions
+set(cppzmq_INCLUDE_DIR ${fletch_BUILD_INSTALL_PREFIX}/include CACHE PATH "" FORCE)
```
dstoup: Look at some of the existing External_* files. There is a piece about fletch_ROOT that needs to be set for all packages.

CMake/External_cppzmq.cmake
```diff
@@ -0,0 +1,32 @@
...
+ExternalProject_Add(cppzmq
+  PREFIX ${fletch_BUILD_PREFIX}
+  GIT_REPOSITORY ${cppzmq_git_repo}
+  GIT_TAG ${cppzmq_git_tag}
+  PATCH_COMMAND ${CMAKE_COMMAND}
+    -Dcppzmq_patch:PATH=${fletch_SOURCE_DIR}/Patches/cppzmq
+    -Dcppzmq_source:PATH=${fletch_BUILD_PREFIX}/src/cppzmq
+    -P ${fletch_SOURCE_DIR}/Patches/cppzmq/Patch.cmake
+  CMAKE_GENERATOR ${gen}
+  CMAKE_ARGS
```
dstoup: Fletch has a COMMON_CMAKE_ARGS, or something like that , which needs to go for all packages. Have a look at one of the existing CMake builds in fletch to find the variable.

CMake/fletch-tarballs.cmake
```diff
@@ -553,6 +553,20 @@ set(qtExtensions_md5 "928fcd80b6c7e5534bf23c3b2fd76363")
...
 
+if (NOT WIN32)
+# ZeroMQ
+set(ZeroMQ_version "3.2.2")
+set(ZeroMQ_url "https://archive.org/download/zeromq_${ZeroMQ_version}/zeromq-${ZeroMQ_version}.tar.gz")
+set(ZeroMQ_md5 "df68431d9300da84a1a5f9a2784e33de")
+list(APPEND fletch_external_sources ZeroMQ)
+
+# CPP ZeroMQ header -- force-builds because of Git source
+set(cppzmq_git_repo "https://github.com/zeromq/cppzmq.git")
```
dstoup: Let's use tag 4.2.3 if it plays nice with ZeroMQ. I would prefer that we stay away from hashes whenever possible.

Also, the  URL should be https://github.com/zeromq/cppzmq/archive/v${cppzmq_version}.zip

Patches/cppzmq/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cmake_minimum_required(VERSION 2.8)
```
dstoup: We probably don't need this patch anymore. The latest version provides a CMakeLists.txt file. Let's start there. If it doesn't work, we can patch it and try to push fixes upstream.

Patches/cppzmq/Patch.cmake
```diff
@@ -0,0 +1,9 @@
+#+
+# This file is called as CMake -P script for the patch step of
+# External_cppzmq.cmake to create a nested directory location for the zmq.hpp
+# header file.
```
dstoup: No patch means this one can go away too.

Patches/cppzmq/CMakeLists.txt
Cookt2: I removed the patch and ran into this build error:

CMake Error at /usr/share/cmake-3.10/Modules/WriteBasicConfigVersionFile.cmake:40 (message):
  No VERSION specified for WRITE_BASIC_CONFIG_VERSION_FILE()
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/CMakePackageConfigHelpers.cmake:207 (write_basic_config_version_file)
  CMakeLists.txt:53 (write_basic_package_version_file)

-- I assume this error is the reason the patch is there; so I kept the patch in the next commit.

CMake/External_ZeroMQ.cmake
```diff
@@ -0,0 +1,21 @@
+fletch_Require_Make()
```
dstoup: Again, zeroMQ does have a CMakeLists.txt file. We should use that rather than configure/make.

CMake/External_cppzmq.cmake
```diff
@@ -0,0 +1,38 @@
...
+   -Dcppzmq_patch:PATH=${fletch_SOURCE_DIR}/Patches/cppzmq
+   -Dcppzmq_source:PATH=${fletch_BUILD_PREFIX}/src/cppzmq
+   -P ${fletch_SOURCE_DIR}/Patches/cppzmq/Patch.cmake
+
+  CMAKE_GENERATOR ${gen}
+  CMAKE_ARGS
+    ${COMMON_CMAKE_ARGS}
+    -DCMAKE_INSTALL_PREFIX:PATH=${fletch_BUILD_INSTALL_PREFIX}
+    -DBUILD_SHARED_LIBS:BOOL=${BUILD_SHARED_LIBS}
+  BUILD_IN_SOURCE 1
```
dstoup: You can remove the BUILD_IN_SOURCE flag. It defaults to 0, which is more appropriate for a CMake build.

CMake/External_ZeroMQ.cmake
```diff
@@ -0,0 +1,21 @@
...
+  INSTALL_DIR ${fletch_BUILD_INSTALL_PREFIX}
+  INSTALL_COMMAND ${MAKE_EXECUTABLE} install
+)
+
+set(ZeroMQ_ROOT ${fletch_BUILD_INSTALL_PREFIX} CACHE PATH "")
+file(APPEND ${fletch_CONFIG_INPUT} "
+#######################################
+# ZeroMQ
+#######################################
+set(ZeroMQ_ROOT @ZeroMQ_ROOT@)
```
dstoup: You have to set `set(ZeroMQ_ROOT   \${fletch_ROOT})`

in the fletch config.

CMake/fletch-tarballs.cmake
```diff
@@ -553,6 +553,20 @@ set(qtExtensions_md5 "928fcd80b6c7e5534bf23c3b2fd76363")
 set(qtExtensions_dlname "qtExtensions-${qtExtensions_version}.zip")
 list(APPEND fletch_external_sources qtExtensions)
 
+if (NOT WIN32)
```
dstoup: Why block this build on Windows?

CMake/External_cppzmq.cmake
```diff
@@ -0,0 +1,38 @@
+# This repo consists of a single hpp header file that wraps the ZeroMQ header.
+# If someone wants this, they probably want ZeroMQ also, however this is not a
+# dependency to "install" the cppzmq header. If ZeroMQ is not enabled, a
+# warning is shown.
+
+if(NOT fletch_ENABLE_ZeroMQ)
+  message(WARNING "cppzmq module enabled without ZeroMQ! Only the header will be installed!")
+endif()
+
+ExternalProject_Add(cppzmq
```
dstoup: This build is missing the DOWNLOAD_DIR. It is dropping the zip in the wrong directory. Also, you should use the download_name (I think that's what it is, see other packages for example) to rename the zip file to something that is recognizable as cppzmq. Right now, it's just v4.2.3.zip

Issue #159, opened 2017-07-28
fletch cmake minimum version < than cmake minimum version of some packages


d-hoke:
Fletch readme and CMakeLists.txt indicates minimum cmake version of 2.8.12.

Some of the packages built by fletch currently appear to need minimum of cmake 3.1 for some host environments.

(grep -r --include="CMakeLists.txt" cmake_minimum_required . from top of fletch binary dir.)

dstoup:
Fixed in https://github.com/Kitware/fletch/pull/291

PR #431, opened 2018-08-13
Increase min cmake version and set it higher if VXL built with cxx11

dstoup:
VXL with cxx11 requires CMake version 3.9.5, otherwise it requires 3.3. This branch makes the minimum CMake version for fletch 3.3 since several packages require that. It also conditionally increases that to 3.9.5 if VXL and cxx11 are enabled.


dstoup:
fixes issue #159

PR #438, opened 2018-08-30
Fix Boost build for MVSC 15.8.2

zackgalbreath:
b2 started failing after the upgrade to MVSC 15.8.2.
The error was:
  `'cl' is not recognized as an internal or external command, operable program or batch file.`
Updating the MSVC toolset version passed to b2 fixes this problem.


Cookt2:
Jenkins test this please

zackgalbreath:
Nevermind, this won't work.  It changes the resulting Boost library filenames such that CMake doesn't know how to find them anymore.  A more robust solution to this problem is outlined in #439.

PR #442, opened 2018-09-17
Update the ITK version for it's vnl support of newer compilers

dstoup:


wphicks:
LGTM once it passes CI.

PR #453, opened 2018-10-04
Disable Qt5 module qtlocation for all platforms

chetnieter:
The qtlocation module is currently disabled for Qt5 for Windows. I am seeing build errors on both Linux and macOS as well so I would like to disable it for all platforms.

### DIFF
--- a/CMake/External_Qt.cmake
+++ b/CMake/External_Qt.cmake
@@ -222,6 +222,13 @@ list( APPEND Qt_configure
   ${Qt_args_framework}
   )
 
+if (NOT Qt_version VERSION_LESS 5.0.0)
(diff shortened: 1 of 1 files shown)


chetnieter:
@dstoup PTAL

Cookt2:
Since there is no way to test Qt5 with our PR dashboards yet, I'm currently running these builds targeted at your source branch for this PR: 
http://proteus:8080/job/FletchMacNightlyPy3Qt5/ 
http://proteus:8080/job/FletchLinuxNightlyPy3Qt5/
I will update this PR with the results once they are done. In the future, hopefully we can get more machines to cover all(or most) options for building Fletch.

dstoup:
This fix looks good regardless of current dashboard status.

PR #458, opened 2018-10-15
Support python debug libs + python lib installs

aaron-bray:
If one chose to install python' debug libraries (which are required to build debug fletch with opencv on windows) python discovery returns a list for python libraries. we need to parse that list and store libraries to the appropriate variables

Also, By default, OpenCV installed its python library to the site-packages where fletch found python, which may be Program Files where the build cannot write

We need to make a consistent location in the fletch build for libraries to put their build python packages. This PR has that for OpenCV. Note for some packages (like Caffe) there is no way to direct where their build puts their built python packages

### DIFF
--- a/CMake/External_OpenCV.cmake
+++ b/CMake/External_OpenCV.cmake
@@ -202,6 +202,22 @@ else()
   list(APPEND OpenCV_EXTRA_BUILD_FLAGS -DBUILD_JPEG=ON)
 endif()
 
+set(OpenCV_PYTHON_FLAGS 
+     -DBUILD_opencv_python2:BOOL=OFF
+     -DBUILD_opencv_python3:BOOL=OFF
+   )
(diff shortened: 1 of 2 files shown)


dstoup:
Jenkins test this please

dstoup:
Quick question. Does this patch require the python_debug libraries to be installed? In other words, will it also work for users who want a Debug build with Python enabled but don't have any need to install Python's debug libraries?

aaron-bray:
No, it does not require debug to be installed.
But if you do install debug libs, you need this to build both release and debug

It goes off the Python Lib string returned by cmake, it tests to see if its a list (by len)
if it is a list, I try to parse it

dstoup:
Sounds good, thanks. I will do some Windows testing in addition to the dashboards just to poke at it some.

PR #473, opened 2018-11-08
Disable building the VTK tests

mleotta:
Fletch does not run the package tests anyway, so this will save build time.


mleotta:
@dstoup I think the dashboards passed as much as can be expected.

dstoup:
> @dstoup I think the dashboards passed as much as can be expected.

Yes they did ... merging. Thanks!

PR #479, opened 2018-12-28
Update VXL and ITK

mattdawkins:
VXL update was required for the ITK version


mattdawkins:
You can...  talk?

dstoup:
If this patch fixes the upstream VXL's ffmpeg issue then please push it there instead of making a fletch patch. I can approve and merge VXL branches so I would prefer not patching locally.

dstoup:
jenkins test this please

Patches/VXL/vidl_ffmpeg_ostream_v56.hxx
```diff
@@ -0,0 +1,606 @@
+// This is core/vidl/vidl_ffmpeg_ostream_v56.hxx
```
dstoup: Is this a new file or patched from VXL source? If it's patched, it would be ideal to add in 2 commits so we can see the actual diff

dstoup:
Having looked at the diff in the VXL patch, is there a good reason we don't push this into that repo? Also, why don't we go closer to master? Is something else broken?

mattdawkins:
Something else was broken on latest VXL in the time period of this branch though that's fairly old now so maybe it was fixed, haven't had a chance to look at this in more detail and this branch is still broken on Windows due to a bug in VXL's build sys

dstoup:
Jenkins test this please

dstoup:
I have been testing recently and currently master fails on VS with link errors.

PR #482, opened 2019-01-23
Patch OpenCV_contrib MacOS linker issue

mleotta:
This was a problem with OpenCV 3.4.0 on MacOS using freetype from macports

See also upstream PR here
https://github.com/opencv/opencv_contrib/pull/1972


Cookt2:
Jenkins test this please

dstoup:
I think the errors that are coming from building kwiver against this branch are being caused by a path too long issue only. Increases my love of Windows. I believe it's ok to merge this anyway.

PR #483, opened 2019-01-25
Fix for GDAL picking up macport libkml over Fletch libkml on Apple

Cookt2:
If Fletch is building libkml, use it. If Fletch is not building libkml, try to use macport/system libkml.

### DIFF
--- a/CMake/External_GDAL.cmake
+++ b/CMake/External_GDAL.cmake
@@ -67,9 +67,15 @@ else()
     #
     # Also set '--with-local=/usr' to force selection of /usr/lib/libiconv over e.g. /opt/local/lib
     # from macports.  GDAL's '--with-libiconv-prefix' option looks like it should handle
-    # this but in fact seems to do nothing.
-    #
-    set(_GDAL_ARGS_APPLE --without-libtool --with-netcdf=no --with-curl=no --with-local=/usr)
+    # this but in fact seems to do nothing. 
+
+    if(fletch_ENABLE_libkml)
+      #If we're building libkml on apple, then use it. 
+      list(APPEND _GDAL_DEPENDS libkml)
+      set(_GDAL_ARGS_APPLE --with-libkml=${KML_DIR} --without-libtool --with-netcdf=no --with-curl=no --with-local=/usr)
+    else()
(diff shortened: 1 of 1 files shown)


Cookt2:
This should be targeted to release. I'm going to let the dashboards run, and if accepted I'll target release.

mleotta:
This does not fix the problem for me.  I get the same error.  I've tried tweaking your change a bit, but no luck so far.

CMake/External_GDAL.cmake
```diff
@@ -67,9 +67,15 @@ else()
...
     # Also set '--with-local=/usr' to force selection of /usr/lib/libiconv over e.g. /opt/local/lib
     # from macports.  GDAL's '--with-libiconv-prefix' option looks like it should handle
-    # this but in fact seems to do nothing.
-    #
-    set(_GDAL_ARGS_APPLE --without-libtool --with-netcdf=no --with-curl=no --with-local=/usr)
+    # this but in fact seems to do nothing. 
+
+    if(fletch_ENABLE_libkml)
+      #If we're building libkml on apple, then use it. 
+      list(APPEND _GDAL_DEPENDS libkml)
```
mleotta: We probably want to add this dependency for all platforms, not just Apple.

CMake/External_GDAL.cmake
```diff
@@ -67,9 +67,15 @@ else()
...
+      set(_GDAL_ARGS_APPLE --with-libkml=${KML_DIR} --without-libtool --with-netcdf=no --with-curl=no --with-local=/usr)
```
mleotta: I think `KML_DIR` here needs to be `KML_ROOT` instead (or just `fletch_ROOT`) because `KML_DIR` points to the CMake exports file, not the install root. 

This change almost works.  It finds the headers, but not the library, and marks libkml as not found and then I get the same error as before.  I only have "libkml.a", maybe it's looking for the dynamic library?

mleotta:
by the way it's `libxml2` causing the problem I am seeing, not `libkml`

mleotta:
I've made changes on #484 that seem to address the issues I was seeing.

Cookt2:
Closing in favor of #484

PR #484, opened 2019-01-25
Add GDAL dependency on libxml2

mleotta:
This change seems to fix my issue on macOS.  It seem like it might be relevant on other platforms too.


Cookt2:
Do we want to target release with this fix?

mleotta:
I suppose the release branch makes sense for this.  I will rebase it (and add a release note).  But it might make sense to let the current CI builds finish just to confirm that this is not problematic on master.

dstoup:
Looks like the Windows dashboard went down at some point. Restarting it now and will run that dashboard and merge.

dstoup:
I'm just rerunning the dashboard to test the build_kwiver post task to make sure it's fixed now. I will merge soon.

PR #486, opened 2019-02-01
Add PDAL

mleotta:
I need to add PDAL for point cloud I/O support (e.g. LAS files).  I haven't been able to get this to work yet.  I'm getting linking errors.  I could use some help.


Cookt2:
Linux build failure is not related to this branch. Running the dashboards again.

Cookt2:
Jenkins test this please

mleotta:
I was able to work around the Linux error locally by setting LD_LIBRARY_PATH to the Fletch install/libs. I'm not sure the right way to fix this in CMake.  It may actually be an error in how we build GDAL.  The GDAL library does not have a path to png.

Cookt2:
Jenkins test this please

dstoup:
I have found GDAL a bit flaky as well about its treatment of internal packages for its build. In CMake we usually do something like, e.g. in GDAL

    set(env ${CMAKE_COMMAND} -E env)
    set(env_var LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${fletch_BUILD_INSTALL_PREFIX}/lib)
    set(GDAL_CONFIGURE_COMMAND ${env} ${env_var} ${GDAL_CONFIGURE_COMMAND})

which works for Linux. There is likely a better way to handle this, but that's what we've done in the past.

dstoup:
Otherwise, I think the code looks good.

Cookt2:
Jenkins test this please

Cookt2:
There are errors when PDAL is built without GEOS enabled. GEOS should be added as a required dependency for PDAL. I did that here: https://github.com/Kitware/fletch/commit/9bd1f538a326dbe84622d62014d7cfadfdf8e06f

I'm currently looking into why Windows is not happy with this branch.

Cookt2:
From PDAL configure step:
```
-- The following REQUIRED packages have been found:

 * GDAL (required version >= 1.9.0)
   Provides general purpose raster, vector, and reference system support
 * GEOS (required version >= 3.3)
   Provides general purpose geometry support
 * GeoTIFF
 * ZLIB
   Compression support in BPF
 * Threads, The thread library of the system
```

mleotta:
The Windows builds seem to have errors that are not related to this branch, and the non-GPU Windows machines did not re-run after the latest changes.

Cookt2:
Jenkins test this please

Cookt2:
I've created #500 that addresses some issues I had with this branch.

mleotta:
Can this be merged now?  If not, what changes are needed?

Cookt2:
This looks good to me, the linuxGPU failures aren't related to this branch.

PR #488, opened 2019-02-04
Add ffmpeg 3.3.3 option to windows

chetnieter:
ffmpeg version 3.3.3 is now an option on windows. Just like the 2.6 version it is downloading pre-built binaries.


mleotta:
Jenkins test this please

chetnieter:
@dstoup We are already aware of some issues with VXL on windows with 3.3.3 that show up in the KWIVER tests.

dstoup:
Got it, thanks! I'll go ahead an merge this branch as is then. I did push the other branch anyway, and will just let it finish.

PR #497, opened 2019-02-13
Fix typo in CMake symbol name.

linus-sherrill:
This type causes YAML to be downloaded even if it has already been done.


dstoup:
Thank you and great find!

PR #499, opened 2019-02-13
Add feature request template

Cookt2:
This will be the template that is suggested when submitting a feature request.


Cookt2:
I'm going to merge these templates. They aren't groundbreaking, and can always be adjusted later if needed.

PR #500, opened 2019-02-14
add-PDALv2

Cookt2:
This builds on #486. 

Windows needed an extra CXX flag "/D_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING" to build, was getting `error C2220: warning treated as error - no 'object' file generated`
related: https://github.com/google/googletest/issues/1373 

Windows also needed the locations of libs/includes for Fletch built dependencies.

GEOS needed to be added to the list of required dependencies (I was seeing build errors when not using fletch_enable_all_packages=on)


Cookt2:
Bulk of this code has been reviewed already on #486, except for the last 3 commits.

mleotta:
It might be better to make a pull request into my branch (on my fork) instead.  That way it will update the original PR.

PR #502, opened 2019-02-18
[WIP]Update libkml to version 1.3.0

Cookt2:
This changes libkml source from an old Kitware fork hash to libkml/libkml offical release.

Updating libkml to 1.3.0 allows GDAL to build with proper libkml support.

Libkml 1.3.0 requires ZLib and boost as dependencies.

This PR depends on #503. 

See: https://github.com/libkml/libkml/releases/tag/1.3.0

PR #503, opened 2019-02-18
update ZLib to 1.2.9, fix ZLib find errors

Cookt2:
Prerequisite to #502 

Update ZLib from version 1.2.8 (or 1.2.3 according to the CMakeLists.txt) to version 1.2.9, which is required to build libkml 1.3.0 for this PR #502.

This also changes the ZLib source from a Kitware fork to an official ZLib release. (Injecting our custom CMakeLists.txt to build it)

...as well as a bonus ZLib library find fix for Apple

### DIFF
--- a/CMake/External_PNG.cmake
+++ b/CMake/External_PNG.cmake
@@ -8,7 +8,7 @@ add_package_dependency(
 )
 if(NOT ZLIB_FOUND)
   set(ZLIB_INCLUDE_DIRS ${fletch_BUILD_INSTALL_PREFIX}/include)
-  get_system_library_name(z zlib_lib)
+  get_system_library_name(zlib zlib_lib)
   set(ZLIB_LIBRARIES ${fletch_BUILD_INSTALL_PREFIX}/lib/${zlib_lib})
 endif()
 set(png_cmake_args ${png_cmake_args}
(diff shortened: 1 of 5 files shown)


Cookt2:
Jenkins test this please

Cookt2:
LinuxGPU failures are unrelated to this branch, this looks ready to me.

Cookt2:
> My first and really only critique is, it would be good if the patch commit was split into 2. It is helpful if you add the file on one commit and apply the changes on a separate commit so we can see what you did. In this case though, maybe your patch is just the whole CMakeLists.txt write?
> 
In the future I will do this. I copied the CMakeLists.txt from: https://github.com/commontk/zlib/blob/master/CMakeLists.txt (This is from the Kitware fork we were using), and the only change was:

SET(ZLIB_VERSION_PATCH "3") -> SET(ZLIB_VERSION_PATCH "9")

> The other comment, which is possibly done on a followup branch is, I have never really liked that we created output libraries that don't match the standard zlib naming. It has caused issues in the past. Ideally, our library should be named the same as a system installed zlib on all platforms.

I definitely agree with you on this. I think we should switch to using the CMakeLists.txt that is shipped with ZLib (which names the library correctly). Injecting our custom CMakeLists.txt file worked this time around, but if they bump a minor or major version, it is less likely to work. I can do this work in a follow-up PR.

PR #510, opened 2019-03-18
Force gtest to build c++ 98 for gcc less than 4.9

dstoup:


dstoup:
The failure building kwiver is not related to this branch. It's a recent kwiver regression which I'm investigating in parallel.

PR #514, opened 2019-04-05
update dockerfile to support remote builds

based64god:
Also updates the version of cmake and ubuntu that we build against. It might be worth migrating the base image to an nvidia one.

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -1,6 +1,6 @@
 # Fletch Dockerfile
-
-FROM ubuntu:trusty
+ARG UBUNTU_VER=18
+FROM ubuntu:${UBUNTU_VER}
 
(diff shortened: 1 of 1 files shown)


based64god:
PTAL @Cookt2

dockerfile
```diff
@@ -1,6 +1,6 @@
 # Fletch Dockerfile
-
-FROM ubuntu:trusty
+ARG UBUTNU_VER=18
+FROM ubuntu:${UBUTNU_VER}
```
Cookt2: ubutnu -> ubuntu?

Cookt2:
Otherwise looks good to me

dockerfile
based64god: whoops

PR #518, opened 2019-04-16
Patch GDAL

m-chaturvedi:
Adding a few more TREs.
`nitf_spec.xml` from `2.3.2`:
https://raw.githubusercontent.com/OSGeo/gdal/v2.3.2/gdal/data/nitf_spec.xml

### DIFF
--- a/Patches/GDAL/2.3.2/Patch.cmake
+++ b/Patches/GDAL/2.3.2/Patch.cmake
@@ -1,0 +1,10 @@
+#+
+# This file is called as CMake -P script for the patch step of
+# External_GDAL.cmake.
+# GDAL_patch and GDAL_source are defined on the command line along with the
+# call.
+#-
+
+message("Copying ${GDAL_patch}/data/nitf_spec.xml to ${GDAL_source}/data/nitf_spec.xml")
+file(COPY ${GDAL_patch}/data/nitf_spec.xml DESTINATION ${GDAL_source}/data/)
+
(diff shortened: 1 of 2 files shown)


m-chaturvedi:
+@dstoup for review, please.

Patches/GDAL/2.3.2/Patch.cmake
```diff
@@ -0,0 +1,12 @@
+#+
+# This file is called as CMake -P script for the patch step of
+# External_GDAL.cmake.
+# GDAL_patch and GDAL_source are defined on the command line along with the
+# call.
+#-
+
+# Fix the version number to not contain "dev" and add a SOVERSION
+# for the public C API
+message("Copying ${GDAL_patch}/frmts/mrsid/nmake.opts to ${GDAL_source}/frmts/mrsid")
```
dstoup: This message seems wrong ... copy/paste from a different patch file?

dstoup: Everything else looks good. I will run this through the dashboards after this fix. Assuming it passes, I will merge after.

m-chaturvedi: Yes, indeed. 
Removed.

dstoup:
Jenkins test this please

Patches/GDAL/2.3.2/Patch.cmake
dstoup: The comment was a good catch too, but I was speaking about the CMake message

message("Copying ${GDAL_patch}/frmts/mrsid/nmake.opts to ${GDAL_source}/frmts/mrsid")

That isn't correct either.

m-chaturvedi: Sorry.

dstoup: No worries, thanks.

PR #525, opened 2019-05-15
Update FFMPEG default version to 3.3.3

mleotta:
Recently approved change in KWIVER (Kitware/kwiver#799) require newer FFMPEG for fast accurate seeking in the FFMPEG arrow.  Some tests will fail if we use the older version 2.6.2.

The only potential downside is that the old VXL implementation (vidl_ffmpeg) used in KWIVER may not work as reliably with the new FFMPEG.  I suggest we update the default version but continue to support 2.6.2 as an option.  This will force us to either update or abandon vidl_ffmpeg.


mleotta:
@chetnieter may want to comment on this as well.

mleotta:
@dstoup I've discussed this with several people, and the consensus is that we should merge this as long as we keep the fallback option to build with the older version (as this branch does).

Can you merge it?

PR #526, opened 2019-05-17
Set PDAL_DIR

mleotta:
This helps packages using Fletch to provide PDAL find version that is in Fletch.


mleotta:
@dstoup Can we merge this?  The clang error seems to be unrelated.

PR #528, opened 2019-05-29
Help PDAL find dependencies on macos

mleotta:
Macos also needs help finding some of these libraries, not just WIN32.

### DIFF
--- a/CMake/External_PDAL.cmake
+++ b/CMake/External_PDAL.cmake
@@ -14,6 +14,15 @@ endif()
 if (fletch_ENABLE_GDAL)
   message(STATUS "PDAL depending on internal GDAL")
   list(APPEND PDAL_DEPENDS GDAL)
+  if(WIN32)
+    set(GDAL_LIB_NAME "gdal_i.lib")
+  else()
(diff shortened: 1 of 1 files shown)


mleotta:
@dstoup @Cookt2 This change resolves by build issues on macOS in which PDAL was finding several system Frameworks over the Fletch-built libraries.  While this change fixes the problem, I also found similar issues in KWIVER and then came across a more general solution that we might want to also consider for Fletch.  If we
```
set(CMAKE_FIND_FRAMEWORK LAST)
```
before any `find_package` calls it will give Frameworks lower priority than the fletch libraries. 
See also https://github.com/Kitware/kwiver/pull/808

mleotta:
Is the clang CI build still broken regardless of these changes?  Everything else is clear. Can we merge?

dstoup:
> Is the clang CI build still broken regardless of these changes? Everything else is clear. Can we merge?

The Clang build is broken, yes. It won't build fletch correctly on the the CentOS 7 build of clang. I need to move the dashboard to a more modern distro to fix it. It can safely be ignored until then.

mleotta:
This does fix my issues

PR #532, opened 2019-06-05
When enabling mrsid in gdal, explicitly disable openjpeg

dstoup:
If we don't explicitly disable openjpeg, GDAL will find it and include the plugin. If GDAL has both mrsid and openjpeg plugins, it favours openjpeg which is not capable of opening some of our nitf files..


dstoup:
Jenkins test this please

PR #546, opened 2019-07-23
update yasm source link to github, current one is dead and unstable

Cookt2:
This changes the yasm target source from tortall.net to github, I think this will allow for better uptime on the link.


dstoup:
I don't think the mac errors have anything to do with yasm. Merging.

PR #558, opened 2019-10-15
Dev/ffmpeg static build

as6520:
This PR updates the ffmepg build to support static build that is used for kwiver wheel.


dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

dstoup:
The Windows build errors are just the usual 3 tests that 'timeout' on the dashboard for some reason and I believe can be ignored. The MAC errors are real though.

https://open.cdash.org/viewBuildError.php?buildid=6738947

as6520:
Thanks for the info, I will look into it.

dstoup:
Jenkins test this please

as6520:
Jenkins test this please

CMake/External_FFmpeg.cmake
```diff
@@ -52,9 +52,12 @@ else ()
 
   # Should we try to point ffmpeg at zlib if we are building it?
   # Currently it uses the system version.
-  #if (fletch_ENABLE_Zlib)
-  #  list(APPEND ffmpeg_DEPENDS ZLib)
-  #endif()
+  if (fletch_ENABLE_Zlib)
+    list(APPEND ffmpeg_DEPENDS ZLib)
+    set(_FFmpeg_zlib --disable-zlib)
```
dstoup, line 57: It seems like these calls are backwards but it's hard to tell exactly from the config log and help.  My fletch without any changes links to the fletch zlib, which means we probably should add it to ffmpeg_DEPENDS but not sure about the enable/disable calls.

as6520: You are correct the help in the configure is fairly ambigious. However based on https://gist.github.com/YurySolovyov/e8cef2ee6cb895d20057f2ea3a36870e, `--disable-zlib` is supposed to disable zlib that is autodetected. Based on the documentation of the section, only the system libraries are autodetected, however I am not sure why they won't be linked to ffmpeg as the documentation suggests when the flag is on.

dstoup (approved):

PR #561, opened 2019-10-28
Update protobuf3 and use CMake

dstoup:
Currently this branch sets the default Protobuf version to 3 for testing purposes only. We have more work to do before version 3 can be the default.l If this branch passes dashboards, I will pull the version back to 2 and repush.


dstoup:
Jenkins test this please

dstoup:
@aaron-bray The dashboards passed but I think you'd mentioned something about kwiver's definitions needing to change as a result of pb3. Should I keep 2 as the default for now? Thinking that's the right thing anyway until we get buyin, but wanted to get a sense of the rest of the work left to do.

CMake/External_Protobuf.cmake
```diff
@@ -1,43 +1,50 @@
...
+if (Protobuf_SELECT_VERSION STREQUAL "3.9.0")
+  ExternalProject_Add(Protobuf
+#    PREFIX protobuf
+    URL ${Protobuf_url}
+    URL_MD5 ${Protobuf_md5}
+    SOURCE_SUBDIR ./cmake
+    ${COMMON_EP_ARGS}
+    ${COMMON_CMAKE_EP_ARGS}
+#    UPDATE_COMMAND
+#        COMMAND ${CMAKE_COMMAND} -DPULSE_IL2CPP_PATCH=${PULSE_IL2CPP_PATCH} -Dprotobuf_source=${protobuf_SRC} -Dprotobuf_patch=${protobuf_Patch} -P ${protobuf_Patch}/Patch.cmake
```
aaron-bray: You can take this patch out. It was put in to work with the Unity IL2CPP compiler that converts C# to C++.
The IL2CPP compiler can be a bit too aggressive, so this patch was created
By the time you are supporting kwiver/fletch into C# and Unity, this will probably have changed
Just let me know then

dstoup: Thanks ... what are you thoughts about making protobuf3 the default? It certainly works building/testing kwiver on all platforms.

CMake/External_Protobuf.cmake
```diff
@@ -1,43 +1,50 @@
...
+    CMAKE_ARGS
+       ${COMMON_CMAKE_ARGS}
+       -Dprotobuf_BUILD_TESTS:BOOL=OFF
+       -Dprotobuf_BUILD_EXAMPLES:BOOL=OFF
+       -Dprotobuf_BUILD_SHARED_LIBS:BOOL=${BUILD_SHARED_LIBS}
```
aaron-bray: Please read this tid bit about protobuf linkage
https://github.com/protocolbuffers/protobuf/tree/master/cmake#dlls-vs-static-linking

I suggest we always build protobuf as static libs, and they should only be linked into the protobuf arrow and nothing else. Should work great in the kwiver architecture.

aaron-bray:
protobuf3 can read in and use protobuf2 files
I have also built protobuf3 in fletch and used that in a Caffe build, which has protobuf2 proto files, and everything still worked fine.

So you really don't need to move the kwiver proto files fro 2 to 3 immediately
It does not look like protobuf2 supports JSON
And there are other differences between the versions
https://groups.google.com/forum/#!topic/protobuf/h-nwPLb42ac

You should be able to remove protobuf2 altogether from fletch and just use 3
I think we should move to protobuf3 proto files unless there is a compelling reason to stay at 2
But we can do that on our own time and still only build protobuf3



So I feel pretty confident we can nix proto2 from fletch all together and not sink any ships

dstoup:
Perfect, thanks! I will keep protobuf3 as the default, leave protobuf2 for just a little while and look to remove it completely soon. I'm looking into the static thing as well. I suspect I will have to force PIC in those cases since it didn't build correctly out of the box that way (on  Linux). I think VS was fine building static

aaron-bray:
Note, 
I think using the protobuf3 library (even with protobuf2 proto files) we should still get JSON support...
But I would want to test that just to be sure...

PR #562, opened 2019-11-11
Optionally Build Python in Fletch

mattdawkins:


dstoup:
In general, before a full review of this PR, it would be useful to know why we need Cython added. Also, this decision needs a broader audience. I am adding a few but if you think of more, please add.

@kfieldho @mleotta @linus-sherrill @Purg @Erotemic

mattdawkins:
At least one package in fletch and another external one linking against fletch required cython in order to install it (e.g. to get past its setup.py) don't remember which.

I don't feel that spending time squashing commits is a current good use of time, especially when branches have already been merged into others and changing history will cause conflicts.

Erotemic:
@dstoup and @mattdawkins note that this PR is adding [CPython](https://en.wikipedia.org/wiki/CPython), not [Cython](https://en.wikipedia.org/wiki/Cython) which is a different thing.

Also note that while its possible to build a relatively bare-bones version of CPython, you will likely want to ensure you link against openssl, readline, sqlite3, gdbm, and lzma in order to get a fully functional interpreter (otherwise it will crash on certain common imports like shelve or hashlib).

@mattdawkins you probably should do a squash. Maintaining a reasonable git history is important for multi-user projects like this one. You can do a pretty simple one by simply doing a soft reset on the base branch, and then creating a new commit that is the combination of all changes. You can then rebase that diff onto the latest master and fix any merge conflicts to get a clean squashed diff.

mattdawkins:
It adds CPython and Cython:

https://github.com/Kitware/fletch/pull/562/files#diff-80b7b602cf627d800bb61221598726f1R152

The pip install list should be parametized with the above defaults IMO

mattdawkins:
And w.r.t. a squashed history, I know how to do it, I just don't know what problem this solves and it introduces more time expense on my end, because I've merged this branch into others and re-writing the commit history would just generate additional conflicts to resolve there. Also I like seeing the history of how something was developed in case an individual commit needs to be revisited or reverted later. If this was a later-stage 100+ user repository or something with people from multiple institutions contributing to, reviewing and/or using, I would agree with you that smaller commit histories are easier to review, but it isn't.

PR #563, opened 2019-11-18
Add option to enable/disable VTK Python wrappings

Erotemic:
The default VIAME build currently fails when Python wrappings for VTK are enabled in Python 3.7. While it might be possible to fix this in VIAME by updating the default VTK version, an alternative option is to simply disable `VTK_WRAP_PYTHON`. This has the additional benefit of shortening build time. 

I added a new option `fletch_ENABLE_VTK_PYTHON`, which does this. It defaults to "ON", so original behavior is preserved. I modified the warning messages so no warning will happen when fletch_ENABLE_VTK_PYTHON is OFF. 

Lastly, this local block of code has a inconsistency in spacing after "if" statements, so I modified those to make them consistent while I was editing the file.

### DIFF
--- a/CMake/External_VTK.cmake
+++ b/CMake/External_VTK.cmake
@@ -187,24 +187,27 @@ set(vtk_cmake_args ${vtk_cmake_args}
   )
 
 # PYTHON
-if (fletch_BUILD_WITH_PYTHON AND NOT MSVC14 )
+if(fletch_BUILD_WITH_PYTHON AND NOT MSVC14 )
+  option(fletch_ENABLE_VTK_PYTHON "Enable Python wrappings for VTK" ON)
   find_package(PythonInterp)
   find_package(PythonLibs)
 
-  if(PythonInterp_FOUND AND PythonLibs_FOUND)
+  if(PythonInterp_FOUND AND PythonLibs_FOUND AND fletch_ENABLE_VTK_PYTHON)
     set(VTK_WRAP_PYTHON ON)
     message(STATUS "VTK building with python support")
   else()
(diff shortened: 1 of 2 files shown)


CMake/External_VTK.cmake
```diff
@@ -187,24 +187,28 @@ set(vtk_cmake_args ${vtk_cmake_args}
   )
 
 # PYTHON
-if (fletch_BUILD_WITH_PYTHON AND NOT MSVC14 )
+option(fletch_ENABLE_VTK_PYTHON "Enable Python wrappings for VTK" ON)
```
mattdawkins: Only recommendation would be to move this option into the fletch_BUILD_WITH_PYTHON AND NOT MSCV14 block so that it only gets shown to the user if python turned on (otherwise it will add the flag to cmake, but if the user tries to turn it on it will have no effect)

Erotemic: done

mattdawkins:
LGTM

CMake/External_VTK.cmake
```diff
@@ -187,24 +187,27 @@ set(vtk_cmake_args ${vtk_cmake_args}
   )
 
 # PYTHON
-if (fletch_BUILD_WITH_PYTHON AND NOT MSVC14 )
+if(fletch_BUILD_WITH_PYTHON AND NOT MSVC14 )
+  option(fletch_ENABLE_VTK_PYTHON "Enable Python wrappings for VTK" ON)
```
mattdawkins: I would also make the flag be advanced

dstoup:
In the future, please don't merge any branches that haven't run through dashboards.

PR #567, opened 2019-12-11
Force fPIC for YAMLcpp

dmjoy:
Regardless of whether or not we're building fletch as shared or
static, we'll likely want to include YAMLcpp in a shared
object (i.e. Kwiver plugin) so we want to ensure that it's always
built with fPIC


dstoup:
Jenkins test this please.

dstoup:
Thanks, this package has been a problem on Windows for a while now.

Patches/YAMLcpp/CMakeLists.txt
```diff
@@ -179,9 +179,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR
 	#
 	set(GCC_EXTRA_OPTIONS "")
 	#
-	if(BUILD_SHARED_LIBS)
-		set(GCC_EXTRA_OPTIONS "${GCC_EXTRA_OPTIONS} -fPIC")
-	endif()
+	set(GCC_EXTRA_OPTIONS "${GCC_EXTRA_OPTIONS} -fPIC")
```
dstoup: The proper, modern way to do this is 
`set(CMAKE_POSITION_INDEPENDENT_CODE ON)`
which shouldn't require editing of the CMake file. Though that is an option, it might make it more apparent if we add 
`-DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON`
as a CMake argument in CMake/External_YAMLcpp.cmake. That will be less mysterious for a future user who is wondering where it came from.

dmjoy: Ah, I see.  I've pushed up changes to this effect

dstoup:
Jenkins test this please

dstoup:
Jenkins test this please

PR #575, opened 2020-03-03
Make caffe version selectable

linus-sherrill:
Make Caffe version selectable. The options are for either version 1 or version 2, with 2 being the default.


dstoup:
Code looks good. Not sure why it would be failing on Windows and Mac. I will have a quick look on Windows.

dstoup:
Nevermind, it's seemingly being caused by a url change in things coming from gitlab. I will test the new URLs and push the change up to this PR.

PR #583, opened 2020-05-08
Fix compilation error with cuda 10.2

danlipsa:
The error is described in https://github.com/opencv/opencv_contrib/issues/1786
and it is caused by cuda 10 making the following change:

NVIDIA Video Decoder (NVCUVID) is deprecated. Instead, use the NVIDIA Video Codec SDK.


dstoup:
Jenkins test this please

dstoup:
There is an error with Windows, but it's in kwiver and not related to this PR. Thanks Dan!

mleotta:
It looks like this is the only branch landed in master after preparing a release candidate.  Since this is a fix, I think it should go in the release branch.

PistonY:
Hello, how I could use `cv::cudacodec::createVideoReader` with cuda 10.2? How to set cmake?
```shell
-D WITH_NVCUVID=ON 
-D BUILD_opencv_cudacodec=ON
```
Is this right?

danlipsa:
@PistonY fletch does not use the function you mention yet so I could get away with just disabling BUILD_opencv_cudacodec in opencv. You might need to update opencv to a version that knows how to properly include cudacodec in newer versions of cuda ( > 10.2) to be able to include that function.

PistonY:
I've fix this thx.

PR #584, opened 2020-05-19
Fix compilation error with cuda 10.2

mleotta:
The error is described in https://github.com/opencv/opencv_contrib/issues/1786
and it is caused by cuda 10 making the following change:

NVIDIA Video Decoder (NVCUVID) is deprecated. Instead, use the NVIDIA Video Codec SDK.


mleotta:
This was already tested and merged into master in https://github.com/Kitware/fletch/pull/583

I'm now merging into `release`

PR #607, opened 2020-07-31
Patch opencv config vs2019

dstoup:
This patch corrects the OpenCVConfig.cmake file for WIN32 by adding recognition of VS 2019.


dstoup:
> I haven't tested it yet, but it looks good to me.

I haven't built kwiver against it yet but the Fletch Darknet did finally build which is my guide that this is the fix. I does add the correct lines to the OpenCVConfig.cmake to determine OpenCV_RUNTIME. I am going to test kwiver quickly as well, but the dashboard run should handle that regardless.

mleotta:
I've now tested this on VS 2019, and built kwiver against it.  It works for me.

dstoup:
Looks like the Windows fail here is a random gtest fail. Merging.

PR #615, opened 2020-09-25
Turn off PCH for gcc<5 regardless of C++ 11

dstoup:


dstoup:
Changed scope to disable PCH for all GCC since it fails on 4.8 as well.

PR #631, opened 2021-01-14
Build boost b2b in release mode

mattdawkins:
I don't actually know enough about this change (or why it was necessary in the first place) so maybe this branch should be abandoned if there's a reason, but using the set release type from cmake caused no issues on my end so maybe the reason for the forced debug is gone/deprecated?


dstoup:
As I mentioned in email, there was an explicit reason for b2b's build to be Debug. I think there is comment about it but not descriptive enough to figure out why.

Given the failure in the Linux build is Boost related, I have to imagine that it's being caused by this change. Fletch builds fine, but building kwiver got broken. Also since b2b is an internal Boost utility, I don't think we want or need to make this change.

PR #632, opened 2021-01-14
Add ffmpeg option for libx264

mattdawkins:


mleotta:
I'm curious about where the need for this arises.  We work with H.264 video all the time and haven't needed to explicitly enable libx264 before.

Is the reason to make this an option because it require GPL code?

mattdawkins:
DIVE team requested it for a specific codec in VIAME Web

Yes, optional because of GPL

daniel-riehm:
Closed in favor of #703

PR #634, opened 2021-01-25
Patch GDAL to build on OSX (fixes missing strdup() prototype).

collinsr:
This patch updates the _XOPEN_SOURCE macro to 600 so GDAL's libjson
finds strdup(). Via https://bugs.launchpad.net/libvterm/+bug/1638205

### DIFF
--- a/Patches/GDAL/2.3.2/Patch.cmake
+++ b/Patches/GDAL/2.3.2/Patch.cmake
@@ -8,3 +8,6 @@
 message("Copying ${GDAL_patch}/data/nitf_spec.xml to ${GDAL_source}/data/nitf_spec.xml")
 file(COPY ${GDAL_patch}/data/nitf_spec.xml DESTINATION ${GDAL_source}/data/)
 
+message("Copying ${GDAL_patch}/ogr/ogrsf_frmts/geojson/libjson/GNUmakefile to ${GDAL_source}/ogr/ogrsf_frmts/geojson/libjson/GNUmakefile")
+file(COPY ${GDAL_patch}/ogr/ogrsf_frmts/geojson/libjson/GNUmakefile DESTINATION ${GDAL_source}/ogr/ogrsf_frmts/geojson/libjson/)
+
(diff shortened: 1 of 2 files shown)


dstoup:
It looks like the Mac build is getting bit by the change in Qt download. That is fixed in master now. Would be worth bringing this PR up to date with master so we can test again.

I can't explain or understand the Linux failure but I don't believe it's related to this change. Digging in to that one now.

dstoup:
If possible, I would love to get this PR sorted out and landed before we make the patch release. I think that means we should switch the destination branch to 'release' if that doesn't cause too much pain.

mleotta (approved):
Looks good to me.  Though I'm not sure if @dstoup wanted to get #634 redirected to `release` and merged before cutting this release.  The comments indicated that was the plan.

mleotta:
Can we address this one way or the other so that we don't hold up the patch release.

collinsr:
Sorry this slipped past my radar on email-- it looks like this needs two things:

1) Restructure so that the patch is constructed correctly (original file first, then diff)

2) Redirect to release rather than master

I believe I just redirected the base to release for (2) -- I assume there was an implied "Roddy, as the submitter, should be the one do this"; if not, my apologies.

If (1) is still a blocker, let me know-- if so, maybe it would be better to abandon this branch and start over?

mleotta:
@collinsr it looks like this branch needs be rebased onto `release`.  Simply redirecting the merge target to release is pulling in all the other changes on `master`.

The KWIVER patch release has been made now.  It would be good to get this in soon so we can make the Fletch patch release as well.

mleotta:
ping @collinsr

Can you rebase this on release?

collinsr:
@mleotta -- I think I did this correctly? My first rebase --onto ...

mleotta:
@collinsr rebase looks good to me. 

It's not obvious to me why the Windows build is failing.  Maybe @dstoup or @Cookt2 or @johnwparent can take a look.

johnwparent:
@mleotta This failure looks like it's coming from an issue @dstoup  and I had talked about with Sprokit and multiconfig generators, where the Sprokit tests. both python and c++  are not provided a configuration, and so, can't be run.

dstoup:
It looks like the Windows failure is expected (though frustrating and disappointing).

When the Fletch PR CI runs the kwiver check to make sure we didn't break anything two tests almost always fail to 'build'. But in this case, it looks like only one of them, test-ffmpeg-video_input_ffmpeg ... there was an ocv test that always failed previously but that seems to have worked today.

I would say this is good to merge.

PR #640, opened 2021-01-26
Upgrade PDAL to 2.2.0

mleotta:
Test if CI can build PDAL 2.2.0.

I think this may require Curl, which Fletch does not provide.

### DIFF
--- a/CMake/fletch-tarballs.cmake
+++ b/CMake/fletch-tarballs.cmake
@@ -376,7 +376,7 @@ endif()
 list(APPEND fletch_external_sources GDAL)
 
 # PDAL
-set(PDAL_SELECT_VERSION 1.7.2 CACHE STRING "Select the version of PDAL to build.")
+set(PDAL_SELECT_VERSION 2.2.0 CACHE STRING "Select the version of PDAL to build.")
 set_property(CACHE PDAL_SELECT_VERSION PROPERTY STRINGS "1.7.2" "2.2.0")
 
 set(PDAL_version ${PDAL_SELECT_VERSION})
--- a/CMake/fletch-tarballs.cmake
+++ b/CMake/fletch-tarballs.cmake
@@ -376,9 +376,16 @@ endif()
 list(APPEND fletch_external_sources GDAL)
 
 # PDAL
-set(PDAL_version 1.7.2)
+set(PDAL_SELECT_VERSION 1.7.2 CACHE STRING "Select the version of PDAL to build.")
+set_property(CACHE PDAL_SELECT_VERSION PROPERTY STRINGS "1.7.2" "2.2.0")
+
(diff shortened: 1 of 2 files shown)


mleotta:
@Cookt2  How do I see the CI failures?  The used to post a link here, but I only see a link to the successful build.

dstoup:
You can do two different things.

1 - Open https://open.cdash.org/index.php?project=Fletch and look under the Experimental section below to find this build. It should have PR 640 at the end of it.

2 - Sometimes the dashboard details isn't very good so you can click the 'Details' link next to the failing node that you want to view. That brings you into the Jenkins web-site. If you need credentials, I can chat you. On the left side, click 'Console Output' and at the top there is a small link 'full log' which shows everything. Searching for the error here is a bit tougher but you should get a better view of the error this way.

dstoup:
In this case, it seems that curl is required but not found .. not sure why that is.

dstoup:
Considering the hard requirement on the libcurl library and headers I wonder if it makes sense to call find_package(CURL REQUIRED) in the External_PDAL.cmake file so we can alert the user right away about missing packages.

I will see about getting the package installed on the dashboard machines though.

dstoup:
Jenkins test this please

dstoup:
Added libcurl to the failing machines.

mleotta:
curl might be a bigger problem on Windows.

I'm not sure how curl is use in PDAL.  I thought it was supposed to be optional but in 2.2.0 it is required (from the CMake perspective).  We could

1) Patch PDAL to not require curl (if it's not critically needed)
2) Add curl as yet another package Fletch provides
3) Not update PDAL and stick with old versions that do not require curl.

There is not a strong pull on updating PDAL yet.  I was just trying to avoid a minor bug in 1.7.2 that I can work around.

dstoup:
It's going to be impossible to extract libcurl from this version. It's used pretty heavily in vendor/arbiter and from the release notes:

doc/development/release-notes/2.1.0.md:- PDAL is now always built using the arbiter library, which is included in the source distribution.

I think it used to disable arbiter when curl wasn't available but no longer.

I'm a bit hesitant to add libcurl to Fletch but it's certainly an option. I'm curious to see what happens on Windows since that might help make a decision. Good to know there is no big push to upgrade but it is still good to have the option to do so. Regarding the bug, there are upgrade options 1.8.0 and 1.9.0 which might avoid the bug you are having but defer the libcurl decision.

One other issue with 2.1.0 that we're going to face, which comes back to the latest comment I made to the kwiver working group, I think pdal 2.1.0 won't build with gcc 4.8.5, the CentOS 7 default. There are two errors on that build, the second one is that 'put_time’ is not a member of ‘std’. Looking in the iomanip header for that compiler confirms that it wasn't implemented at the time of release. 

Anyway, some things to consider.

PR #643, opened 2021-02-09
Upgrade OpenCV to 4.5.1

mleotta:
Add the latest (as of today) OpenCV version, 4.5.1 since 4.2 is still not recent enough to support CUDA 11.  If 4.5.1 works for everyone we can likely remove 4.2 support.

The last commit makes 4.5.1 the default for dashboard testing purposes.

### DIFF
--- a/CMake/External_Caffe.cmake
+++ b/CMake/External_Caffe.cmake
@@ -103,6 +103,10 @@ else()
 endif()
 
 if(fletch_ENABLE_OpenCV)
+  if(OpenCV_SELECT_VERSION VERSION_GREATER_EQUAL 4.0)
+    MESSAGE(FATAL_ERROR "Caffe requires OpenCV version less than 4.0 "
+                        "but ${OpenCV_SELECT_VERSION} is selected")
+  endif()
   set( CAFFE_OPENCV_ARGS
     -DOpenCV_DIR:PATH=${fletch_BUILD_PREFIX}/src/OpenCV-build
     -DOpenCV_LIB_PATH:PATH=${OpenCV_ROOT}/lib
--- a/CMake/External_Caffe_Segnet.cmake
+++ b/CMake/External_Caffe_Segnet.cmake
@@ -104,6 +104,10 @@ else()
 endif()
 
 if(fletch_ENABLE_OpenCV)
+  if(OpenCV_SELECT_VERSION VERSION_GREATER_EQUAL 4.0)
+    MESSAGE(FATAL_ERROR "Segnet requires OpenCV version less than 4.0 "
+                        "but ${OpenCV_SELECT_VERSION} is selected")
+  endif()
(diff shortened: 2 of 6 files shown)


dstoup:
It looks like these failures are related to protobuf. We don't explicitly pass our protobuf build into OpenCV but we do pass some LIB_DIR stuff to all packages. It's possible that OpenCV is getting confused when building its internal version. Maybe passing ours will fix the issue but it's also possible that our version is too old, I would have to look. On the other hand the need for protobuf looks like its coming from opecv_caffe which we may not need or want anyway.

aaron-bray:
I think you should just add this line:

```-DWITH_PROTOBUF:BOOL=False```

 after this line

https://github.com/mleotta/fletch/blob/dev/upgrade-opencv-to-4.5.1/CMake/External_OpenCV.cmake#L312

By default OpenCV is building Protobuf, which I do not think we need to have it do

If we do need OpenCV to support protobuf, that is a larger task that I'd be happy to talk about

mleotta:
I disabled protobuf as suggested by @aaron-bray

mleotta:
Note: On my VS 2019 build I also needed this change: https://github.com/Kitware/fletch/pull/645

mleotta:
It looks like the error on at least some platforms is "OpenCV 4.x+ requires enabled C++11 support".  Does this mean we just can't upgrade until we upgrade the compilers on the CI machines?  I believe we had decided to bump the minimum compiler versions anyway.

dstoup:
> It looks like the error on at least some platforms is "OpenCV 4.x+ requires enabled C++11 support". Does this mean we just can't upgrade until we upgrade the compilers on the CI machines? I believe we had decided to bump the minimum compiler versions anyway.

I believe all CI machines have C++11 support and Fletch should have CMAKE_CXX_STANDARD set to 11 by default. I can look into the issues here and see what's happening. One thing that I need to do in the short-term is upgrade the CentoOS build to RHEL 8 or similar now that we're committing to gcc newer than 4.8.

dstoup:
I see those messages in the Centos build which we can probably ignore and possibly just disable that machine until it's updated to a newer gcc. I am guessing __cplusplus isnt' set to 201103L until full C++ 11 is implemented, which it is not in 4.8.5.

The other machines are failing from Caffe, Carre_Segnet and Darknet not having respective changes to support OCV 4 (I believe ... looks similar to what we saw in kwiver)

mleotta:
I've set the default OpenCV back to 3.4 and added error message for enabling Caffe, Segnet, and Darknet with OpenCV >= 4.0.  I think if the dashboards are okay with these changes we should merge this and then revisit making OpenCV 4 the default later on after we have time to work through the issues with older compilers and incompatibility with other older packages.

dstoup:
Looks good!

PR #644, opened 2021-02-10
Add readme troubleshooting msvc upgrade boost

dstoup:
This PR adds a new Troubleshooting section to the README and adds an entry to document the solution to a common build error in Boost that happens after upgrading your MSVC version.


dstoup:
It's not likely that we need to wait for the dashboard runs since this is a documentation only PR but I will likely force push the branch to fix issues with the rendering of the documentation.

You can view the change by clicking on the link to the PR above and going to the 'code' page for it. That will show the new Troubleshooting section at the bottom.

mleotta (approved):

PR #645, opened 2021-02-13
Fix MSVC issue with CUDA_HOST_COMPILER

mleotta:


mleotta:
Linux failure still seems unrelated

dstoup:
Yes, it's a PDAL issue. Not sure how that started all of a sudden since PDAL seemed fine last I checked. I think this PR is ok to land and I will investigate the PDAL issue. Many package get confused by libjson and the commonality of the header names with other packages.

PR #647, opened 2021-02-17
VTK 9 uses different CMake options for modules than VTK 8

mleotta:
The CMake options for enabling modules have changed in VTK 9, so setting the old options was no longer enabling Qt support in VTK 9.  Note that QtWebKit is not longer a sub-option, so we can just turn the whole Qt group on.

### DIFF
--- a/CMake/External_VTK.cmake
+++ b/CMake/External_VTK.cmake
@@ -91,16 +91,22 @@ else()
     -DVTK_QT_VERSION:STRING=5
   )
 endif()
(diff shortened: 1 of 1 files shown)


dstoup (approved):

mleotta:
Centos seems to have unrelated test failures.  This PR should not change anything for builds unless they are configured to use VTK 9, which is currently not the default.

dstoup:
VTK turns on automatically for kwiver if it's found. I think you're probably right though. I replicated the CI run on that machine and everything passed. Merging.

PR #652, opened 2021-03-10
Patch vtk8.2 gcc10

dstoup:
Patch VTK 8.2 to fix a build issue when using gcc v10. vtkExodusII complains about "multiple definition of `exodus_unused_symbol_dummy_1'"

The VTK issue is written up here and is merged into master as of a year ago:
https://gitlab.kitware.com/vtk/vtk/-/issues/17774

This patch was taken from Slicer. Details posted here:
https://discourse.slicer.org/t/build-fails-in-vtkexodus-on-linux/12018/5


dstoup:
Considering we have the kwiver support for VTK 9.0 in kwiver and it's targetting the minor release, I wonder if it's time to consider moving the default VTK version to 9.0. Thoughts?

PR #653, opened 2021-03-12
downgrade the gdal version to 1.X in the dockerfile

Cookt2:


Cookt2:
@judajake @dstoup PTAL

mleotta:
what is the motivation for this change?

dstoup:
I don't want this to land in master. It will harm most users to benefit one project for a short period of time. I we need to have fletch options considered in the build, I suggest we change the Dockerfile to have, e.g. : 

`ARG GDAL_VERSION=2.3
-DGDAL_SELECT_VERSION=${GDAL_VERSION}`


and call the build like....

`docker build -t ${optional_name_of_image} --build-arg GDAL_VERSION=1.11.5.....`

That will let you do what you need and not hose everyone else that builds Fletch docker for kwiver (which requires 2.3).

Cookt2:
Ok, closing this- as we don't want this in master. GDAL 1.X is needed currently for VIDTK's CI. Will create a temporary fletch/kwiver image that will live in VIDTK's container registry for now, till the permanent fix is in place.

PR #656, opened 2021-03-19
Patch VXL for VS 2019 compile issue

mleotta:
With the latest updates to Visual Studio 2019, this openjpep2 header in VXL starting causing compilation problems.  Apparently the `lrintf()` function has been provided by Visual Studio since 2013 and the `else` case on `#if !defined(_WIN64)` served to define that function for even older VS compilers.  However, something has changed in the lastest VS 2019 where that code is now conflicting with the VS provided definition.  So the easy fix is to just remove that definition.  This should work for all compilers supported by Fletch.

I'm not sure if this is the right fix for upstream VXL because I don't know the oldest version of VS supported by VXL.  Though I doubt the support VS older than 2013 at this point.

I'm actually not even sure why I'm hitting this problem since I'm building openjpeg2 in fletch too.  Why doesn't VXL use the openjpeg2 from fletch?

### DIFF
--- a/Patches/VXL/Patch.cmake
+++ b/Patches/VXL/Patch.cmake
@@ -6,3 +6,8 @@
 file(COPY ${VXL_PATCH_DIR}/core/vbl/vbl_array_2d.h
   DESTINATION ${VXL_SOURCE_DIR}/core/vbl/
 )
+
+# fixes an issue with duplicate definition of lrintf() on Windows
+file(COPY ${VXL_PATCH_DIR}/v3p/openjpeg2/opj_includes.h
+  DESTINATION ${VXL_SOURCE_DIR}/v3p/openjpeg2/
+)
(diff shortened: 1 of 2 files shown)


dstoup:
Regarding the last question, I think it's just an oversight in the configuration. Openjpeg came to Fletch much later than VXL and it was never added to the VXL config. The thing about VXL and its 3p packages is, I think it would likely go ahead and build openjpeg if one wasn't provided so I think the patch here is the right thing either way, but Fletch's openjpeg should get passed into VXL as well.

Regarding pushing up to VXL. If it were the right fix for upstream, it would be good to push it but we're currently stuck on a somewhat older hash of VXL now because of link errors with Visual Studio which means we would still need to patch it for now.  The issue seems to be from some libraries that aren't exported which causes the link errors. I've fixed several of those and pushed upstream but there are a few that have confounded me.

dstoup:
Patch looks good to me though, so I would say merge once CI passes.

PR #662, opened 2021-04-08
Conditionally update zlib name in get_system_library_name

as6520:
The static build for zlib is called libz.a rather libzlib.a. Thus PR checks if the build is static or dynamic and provides the zlib name based on the type of build

### DIFF
--- a/CMake/External_PNG.cmake
+++ b/CMake/External_PNG.cmake
@@ -8,7 +8,11 @@ add_package_dependency(
 )
 if(NOT ZLIB_FOUND)
   set(ZLIB_INCLUDE_DIRS ${fletch_BUILD_INSTALL_PREFIX}/include)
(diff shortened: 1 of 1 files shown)


johnwparent (approved):

dstoup (approved):
I'm ok with this provided CI passes. I have to check on the static CI to see if it's running.

johnwparent:
Looks like Windows failure is due to the Kwiver cmake discover tests issue, rather than a bug in this PR.

dstoup:
Looks good. Do we want/need this patch for the release? If so, change the target to the release branch, otherwise go ahead an merge.

as6520:
I think the fix is required for the kwiver wheel and does not effect the shared build. So it is not necessary to add it to 1.4.x.

PR #663, opened 2021-04-12
Disable FFmpeg assembler optimizations

johnwparent:
Disable FFMPEG assembler optimizations on static build to allow libavcodec to build with fpic enabled. Necessary for Kwiver wheel building.

### DIFF
--- a/CMake/External_FFmpeg.cmake
+++ b/CMake/External_FFmpeg.cmake
@@ -75,6 +75,7 @@ else ()
                            --enable-pic
                            --extra-cflags=-fPIC
                            --extra-cxxflags=-fPIC
+                           --disable-asm
        )
   endif()
   set(FFMPEG_CONFIGURE_COMMAND


dstoup:
I'm ok with the change presented but fair warning, disabling asm might make FFmpeg pretty unusable. That warning comes from FFmpeg when one tries to disable asm.

It's not a big deal either way, but if you're disabling asm in the static build, you probably don't need to include yasm for it, but don't think it will cause harm so probably ok to leave if everything is working as is.

dstoup:
Jenkins test this please.

johnwparent:
If disabling the assembly optimizations may result in usability issues down the line, then I can look for an alternative to this approach, I'd hate to introduce a potential future point of friction.

dstoup:
> If disabling the assembly optimizations may result in usability issues down the line, then I can look for an alternative to this approach, I'd hate to introduce a potential future point of friction.

For sure. I think it's worth merging now anyway so we can get the wheel working and get people using it so we know for sure. We can address any pain points once it's out in the wild.

dstoup:
Jenkins test this please.

dstoup:
I can't tell by looking if the CI ran or not. If it does and passes (which is should) then we can merge.

johnwparent:
Sounds good to me!

dstoup:
Jenkins test this please

PR #668, opened 2021-04-15
WIP: Add gdal 3.2.1

Erotemic:
I need to be able to build gdal 3.2.1 without sudo. I tried to do it standalone, but it is dependency hell. It would be nice if fletch could support this. Hopefully this won't be too difficult to add support for.


dstoup:
Good call ... I was going to set out to do this PR soon. Thanks!

dstoup:
I've made two PR's targeting master that will allow GDAL v3 to build as the default. Once they land, I will push up a quick PR to test GDAL 3 as default, then we should merge this. Eager to get it landed as an option.

PR #673, opened 2021-04-20
GDAL v3 doesn't support a path as config to xml2

dstoup:
This patch is a preparation for GDAL v 3. It won't have any impact on master but will allow GDAL 3 to build properly once that branch lands.


dstoup:
Jenkins test this please

PR #674, opened 2021-04-21
Actually apply VXL patches

mleotta:
It turns out that my previous patches to fix the VXL build were never actually being applied.  I had locally applied them so I didn't notice until a did a clean build.

This also means that the previous existing patch for VXL was never applied either.  Maybe it's not longer needed?  Maybe we should just update VXL from upstream?  I think my patch was fixed upstream, but in a different way.

### DIFF
--- a/CMake/External_VXL.cmake
+++ b/CMake/External_VXL.cmake
@@ -90,6 +90,16 @@ elseif(WIN32)
     )
 endif()
 
+# If a patch file exists, apply it
+set (VXL_patch ${fletch_SOURCE_DIR}/Patches/VXL)
+if (EXISTS ${VXL_patch})
+  set(VXL_PATCH_COMMAND ${CMAKE_COMMAND}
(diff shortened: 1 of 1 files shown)


dstoup:
> It turns out that my previous patches to fix the VXL build were never actually being applied. I had locally applied them so I didn't notice until a did a clean build.
> 
> This also means that the previous existing patch for VXL was never applied either. Maybe it's not longer needed? Maybe we should just update VXL from upstream? I think my patch was fixed upstream, but in a different way.

Attempting to update to something close to upstream master is a great idea. I've picked at it a lot over the last year but haven't been able to get it to build with Visual Studio yet. There are a bunch of link errors. Mostly they are caused by missing exports, which I have fixed and pushed upstream, but there are a few remaining failures that I haven't been successful in fixing. I might take another crack at it soon though because we're getting pretty far from master at this point.

dstoup (approved):
Pending dashboards. Windows CI is moving really slowly and somewhat backed up ATM.

PR #686, opened 2021-05-13
WIP: Update vxl master

dstoup:
This patch updates VXL to today's master, hash 51558266cc28ffb522ae72eebc883a88124a4b6c.

I've updated all patches to reflect upstream changes, including removing the openjpeg patch which isn't needed any more since the issue has been resolved. I've also patched the vnl/io CMakeLists.txt to remove builds of vnl_io_real_polynomial and vnl_io_real_npolynomial on WIN32 platforms. It has a difficult link error but no packages, VXL included, use the io of those modules.


dstoup:
FYI, I've marked this WIP for now. I have it building by itself and kwiver should build with almost all options on (some I can't build on Windows regardless) but I still need to make changes so vidtk will build. Plus, I'm assuming there is a fair amount of testing we should do to make sure the VXL master still behaves the same as our previous hash.

PR #691, opened 2022-02-24
WIP: Version updates for Eigen, GDAL, and FFMpeg

chetnieter:
Updated versions for Eigen, GDAL, and FFMpeg to SWAP compliant versions. The Eigen update is very minor and will be the new default. The updated versions of GDAL and FFMpeg are set to be optional.


mleotta:
We should add a comment in release notes about these updated versions.

There may also be a need to split this into a PR for each package update, but I'll let @dstoup comment on that.  Since only one of these is turned on by default, it might not cause any problems for the dashboards.  However, I think in the past, for new version options, we've added a temporary commit to enable to new version just to run through dashboards with that new version.  If we are doing that we probably want a separate PR for each package.

dstoup:
I think it's worth having a temporary commit or PR that enables the newest versions for testing. It's good to know that folks can build the new versions since they're available. The dashboard does ENABLE_ALL so should cover everything.

Otherwise, no objections to this PR.

dstoup (approved):
Approved pending test of new versions.

chetnieter:
Breaking up this MR into three separate MR to better deal with dashboard failures. The new MR's are #695, #696, and #697

PR #692, opened 2022-02-25
Test new versions for b3d

chetnieter:
This is temporary PR to test the updated versions from #691


chetnieter:
Closing as each package is being handled by it's own PR.

PR #693, opened 2022-02-28
Add option to get OpenCV source from PyPI

chetnieter:
Adding option to OpenCV to use source from PyPI for version 4.5.1.

### DIFF
--- a/CMake/fletch-tarballs.cmake
+++ b/CMake/fletch-tarballs.cmake
@@ -308,8 +308,14 @@ if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES OR AUTO_ENABLE_CAFFE_DEPE
   set_property(CACHE OpenCV_SELECT_VERSION PROPERTY STRINGS "3.4.0" "4.2.0" "4.5.1")
 
   set(OpenCV_version ${OpenCV_SELECT_VERSION})
-  set(OpenCV_url "http://github.com/Itseez/opencv/archive/${OpenCV_version}.zip")
-  set(OpenCV_dlname "opencv-${OpenCV_version}.zip")
+  # Pulling from pypi for version 4.5.1
+  if (OpenCV_SELECT_VERSION VERSION_EQUAL 4.5.1 AND fletch_USE_PYPI_OPENCV)
+    set(OpenCV_url "https://files.pythonhosted.org/packages/bb/08/9dbc183a3ac6baa95fabf749ddb531bd26256edfff5b6c2195eca26258e9/opencv-python-4.5.1.48.tar.gz")
+    set(OpenCV_dlname "opencv-python-4.5.1.48.tar.gz")
+  else()
+    set(OpenCV_url "https://github.com/opencv/opencv/archive/${OpenCV_version}.zip")
+    set(OpenCV_dlname "opencv-${OpenCV_version}.zip")
+  endif()
 
   # Expose optional contrib repo when enabling OpenCV version >= 3.x
   if (NOT OpenCV_SELECT_VERSION VERSION_LESS 3.0.0 )
@@ -324,7 +330,11 @@ if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES OR AUTO_ENABLE_CAFFE_DEPE
 
   # Paired contrib repo information
   if (OpenCV_version VERSION_EQUAL 4.5.1)
-    set(OpenCV_md5 "cc13d83c3bf989b0487bb3798375ee08")
+    if (fletch_USE_PYPI_OPENCV)
+      set(OpenCV_md5 "0e178bd601b25a0a1ee0cd1e8c81bec0")
+    else()
+      set(OpenCV_md5 "cc13d83c3bf989b0487bb3798375ee08")
+    endif()
     set(OpenCV_contrib_md5 "ddb4f64d6cf31d589a8104655d39c99b")
   elseif (OpenCV_version VERSION_EQUAL 4.2.0)
     set(OpenCV_md5 "b02b54115f1f99cb9e885d1e5988ff70")
--- a/CMake/External_OpenCV.cmake
+++ b/CMake/External_OpenCV.cmake
@@ -24,6 +24,12 @@ else()
   unset(fletch_ENABLE_OpenCV_FFmpeg CACHE)
 endif()
 
+# Use the pypi tarball for the opencv source code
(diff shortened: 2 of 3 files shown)


CMake/fletch-tarballs.cmake
```diff
@@ -308,8 +308,19 @@ if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES OR AUTO_ENABLE_CAFFE_DEPE
   set_property(CACHE OpenCV_SELECT_VERSION PROPERTY STRINGS "3.4.0" "4.2.0" "4.5.1")
 
   set(OpenCV_version ${OpenCV_SELECT_VERSION})
-  set(OpenCV_url "http://github.com/Itseez/opencv/archive/${OpenCV_version}.zip")
-  set(OpenCV_dlname "opencv-${OpenCV_version}.zip")
+  # Pulling from pypi for version 4.5.1 since it is SWAP approved
+  if (OpenCV_SELECT_VERSION VERSION_EQUAL 4.5.1)
```
mleotta, line 312: You can remove the duplicate else cases if you do
```
if (OpenCV_SELECT_VERSION VERSION_EQUAL 4.5.1 AND fletch_USE_PYPI_OPENCV)
```

CMake/fletch-tarballs.cmake
```diff
@@ -308,8 +308,19 @@ if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES OR AUTO_ENABLE_CAFFE_DEPE
...
+    if (fletch_USE_PYPI_OPENCV)
+      set(OpenCV_url "https://files.pythonhosted.org/packages/bb/08/9dbc183a3ac6baa95fabf749ddb531bd26256edfff5b6c2195eca26258e9/opencv-python-${OpenCV_version}.48.tar.gz")
```
mleotta, line 314: Since this is a special case that only works for 4.5.1, it might be better to just hard code that version in the path on these two lines, especially since that `.48` would not necessarily apply to other versions if this was generalized in the future.

mleotta (changes requested):
I approve of this change, but requests some minor edits.  As odd as it seems, we have a customer with an interest in pulling OpenCV specifically from the PyPI site.

CMake/fletch-tarballs.cmake
```diff
@@ -308,8 +308,14 @@ if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES OR AUTO_ENABLE_CAFFE_DEPE
...
 
   set(OpenCV_version ${OpenCV_SELECT_VERSION})
-  set(OpenCV_url "http://github.com/Itseez/opencv/archive/${OpenCV_version}.zip")
-  set(OpenCV_dlname "opencv-${OpenCV_version}.zip")
+  # Pulling from pypi for version 4.5.1 since it is SWAP approved
+  if (OpenCV_SELECT_VERSION VERSION_EQUAL 4.5.1 AND fletch_USE_PYPI_OPENCV)
+    set(OpenCV_url "https://files.pythonhosted.org/packages/bb/08/9dbc183a3ac6baa95fabf749ddb531bd26256edfff5b6c2195eca26258e9/opencv-python-${OpenCV_version}.48.tar.gz")
+    set(OpenCV_dlname "opencv-python-4.5.1.48.tar.gz")
+  else()
+    set(OpenCV_url "http://github.com/Itseez/opencv/archive/${OpenCV_version}.zip")
```
Erotemic, line 316: Can we swap http with https here?

CMake/fletch-tarballs.cmake
```diff
@@ -324,7 +330,11 @@ if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES OR AUTO_ENABLE_CAFFE_DEPE
 
   # Paired contrib repo information
   if (OpenCV_version VERSION_EQUAL 4.5.1)
-    set(OpenCV_md5 "cc13d83c3bf989b0487bb3798375ee08")
+    if (fletch_USE_PYPI_OPENCV)
+      set(OpenCV_md5 "0e178bd601b25a0a1ee0cd1e8c81bec0")
+    else()
+      set(OpenCV_md5 "cc13d83c3bf989b0487bb3798375ee08")
```
Erotemic, line 336: I'll also note that we probably should be using something like sha256 or sha512 instead of md5 for content validation. Probably out of scope for this, but we should think about making that change.

CMake/fletch-tarballs.cmake
mleotta: If we are going to switch to https, we should also switch to `https://github.com/opencv/opencv/`.  Since Itseez now redirects to opencv on github.

CMake/fletch-tarballs.cmake
```diff
@@ -308,8 +308,14 @@ if (fletch_ENABLE_OpenCV OR fletch_ENABLE_ALL_PACKAGES OR AUTO_ENABLE_CAFFE_DEPE
   set_property(CACHE OpenCV_SELECT_VERSION PROPERTY STRINGS "3.4.0" "4.2.0" "4.5.1")
 
   set(OpenCV_version ${OpenCV_SELECT_VERSION})
-  set(OpenCV_url "http://github.com/Itseez/opencv/archive/${OpenCV_version}.zip")
-  set(OpenCV_dlname "opencv-${OpenCV_version}.zip")
+  # Pulling from pypi for version 4.5.1 since it is SWAP approved
+  if (OpenCV_SELECT_VERSION VERSION_EQUAL 4.5.1 AND fletch_USE_PYPI_OPENCV)
+    set(OpenCV_url "https://files.pythonhosted.org/packages/bb/08/9dbc183a3ac6baa95fabf749ddb531bd26256edfff5b6c2195eca26258e9/opencv-python-${OpenCV_version}.48.tar.gz")
```
mleotta, line 313: Can we also replace the `${OpenCV_version}.48.tar.gz` with `4.5.1.48.tar.gz` in this line?

CMake/fletch-tarballs.cmake
mleotta: Yes, I would say out of scope for this PR.  We used md5 everywhere in Fletch.  So we should not just change it here.

chetnieter:
@mleotta Dashboards pass when original default version is maintained. We get Caffe build failures when enabled but I confirmed that this also happens with 4.5.1 when it is not pulled from PyPI. I suggest we merge unless there are more comments or suggestions.

PR #694, opened 2022-02-28
Temporary MR to test OpenCV built against PyPI source

chetnieter:
Temporary MR to test building with OpenCV using PyPI source. This is related to #693


chetnieter:
Closing since referring commit is being broken up.

PR #697, opened 2022-02-28
Add option for FFMpeg 4.4.1

chetnieter:


chetnieter:
@mleotta Dashboards passed for both 4.4.1 and original version 3.3.3. Unless you any other comments I believe this is ready to merge.

PR #698, opened 2022-03-09
Only enable Qt in VTK if it is present

chetnieter:
Only enable Qt in VTK if Qt is present. The diff is less than it looks since I just wrapped the Qt setup in an `if(QT_FOUND) - endif()`.


dstoup (approved):
Pending CI success

chetnieter:
@dstoup @mleotta Could one of you merge this. Dave approved it and all the checks pass.

PR #699, opened 2022-03-09
Add VTK 9.1

danlipsa:


danlipsa:
@mleotta Please review.

mleotta (approved):
Looks okay to me

mleotta:
@danlipsa Can you add a release note to `master.txt` indicating that the option for VTK 9.1 has been added?

@dstoup do we need to make a temporary commit that makes VTK 9.1 the default to run it through dashboards?  Or can we merge this as long as the builds succeed with the current default, which is not 9.1?

danlipsa:
@mleotta Do you mean Doc/release.txt?

mleotta:
@danlipsa it should be `Doc/release-notes/master.txt`, not `release.txt` and should say something like "Added option to use VTK 9.1".

danlipsa:
@mleotta An unrelated comment on VTK. If VTK operations are slow, some of them can be sped-up by compiling VTK using VTK_SMP_IMPLEMENTATION_TYPE=STDThread. Right now it is Sequential.
Other available options include TBB and OpenMP but these require external libraries.

mleotta:
@danlipsa We use OpenMP in KWIVER.  I don't think that requires installing anything else if you have the appropriate compilers (I think we had trouble with it on Apple).  Is this option only supported for VTK 9.0 or 9.1 or older versions too?  Would you recommend OpenMP over STDThread if OpenMP is available?

danlipsa:
We use OpenMP for non-intel HPC machines. STDThread backend is the newest, so I think OpenMP is a safer bet.

PR #700, opened 2022-03-14
Add patch to write KLV descriptors

daniel-riehm:
Add a patch modifying FFmpeg to write metadata descriptor structures when writing synchronous KLV to a MPEG-TS file. This is to be in compliance with MISB ST1402.

### DIFF
--- a/Doc/release-notes/master.txt
+++ b/Doc/release-notes/master.txt
@@ -10,6 +10,7 @@ Updates since v1.5.0
 
 * Update Eigen from 3.3.7 to 3.3.9
 * Added option to use FFMpeg 4.4.1
+* Patched FFmpeg to write KLV metadata in compliance with MISB standards
 * Update GDAL from 2.3.2 to 2.4.4
 * Added option to get OpenCV source for version 4.5.1 from PyPI
 * Add option to use VTK 9.1.0
--- a/Patches/FFmpeg/4.4.1/Patch.cmake
+++ b/Patches/FFmpeg/4.4.1/Patch.cmake
@@ -1,0 +1,11 @@
+#+
+# This file is called as CMake -P script for the patch step of
+# External_FFmpeg.cmake.
+#-
+
+# Add patch to write KLV descriptors as required by MISB 1402.
+configure_file(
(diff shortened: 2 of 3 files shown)


mleotta:
It's a bit of a pain, but since CMake can't portably do actually patching and we are overwriting complete files, we recommend that patches added to Fletch are made in at least 2 commits.  The first commit should add the file to be patched in its unchanged form.  The second commit should be the patch to this file.  This way we can see what changes were actually made by reviewing the git history.

daniel-riehm:
> It's a bit of a pain, but since CMake can't portably do actually patching and we are overwriting complete files, we recommend that patches added to Fletch are made in at least 2 commits. The first commit should add the file to be patched in its unchanged form. The second commit should be the patch to this file. This way we can see what changes were actually made by reviewing the git history.

Done

mleotta (approved):
This change looks okay to me.  Please also push upstream to see if we can get this change accepted into FFMPEG.

One issue we will face is that any FFmpeg patches will not impact our Windows builds because we use pre-built Windows binaries.  This is all the more reason to get these changes pushed upstream.

It would also be good to make a release note about this change.  And since 4.4.1 is not the current default, we might want to make a temporary commit that makes 4.4.1 the default just to test that CI can actually build this.

daniel-riehm:
> Please also push upstream to see if we can get this change accepted into FFMPEG.

I have done so, though FFmpeg uses an odd mailing-list system for code review instead of GitHub PRs, so hopefully I submitted it correctly.

> It would also be good to make a release note about this change. And since 4.4.1 is not the current default, we might want to make a temporary commit that makes 4.4.1 the default just to test that CI can actually build this.

Is this in fletch or kwiver?

mleotta:
I was suggesting a release note and temporary commit to trigger CI here in Fletch, not KWIVER.

daniel-riehm:
jenkins test this please

mleotta:
jenkins test this please

daniel-riehm:
The CI is failing only because the mvg tracking test is timing out. Is this something I need to worry about?

PR #702, opened 2022-03-17
Make VTK 9.1 the default

mleotta:


mleotta:
This looks good to merge.  Any final concerns with making VTK 9.1 the default?  Older versions remain options.

PR #721, opened 2022-11-18
Patch FFmpeg to write unregistered SEI user data

daniel-riehm:
MISB ST0604 states that MISP frame timestamps should be put in the "SEI unregistered user data" field of H.264 and H.265 video streams. Our most recent version of FFmpeg, 4.4.1, implements _reading_ of these fields but not writing. This PR pulls in more recent feature additions from the public FFmpeg repository, adding only the code necessary to write these fields (or, more accurately, to tell x264, x265, and nvenc to write these fields).

### DIFF
--- a/Doc/release-notes/master.txt
+++ b/Doc/release-notes/master.txt
@@ -11,7 +11,7 @@ Updates since v1.5.0
 * Update Eigen from 3.3.7 to 3.3.9
 * Added option to use FFMpeg 4.4.1
 * Patched FFmpeg to write KLV metadata in compliance with MISB standards
+* Patched FFmpeg to write unregistered SEI user data
 * Update GDAL from 2.3.2 to 2.4.4
 * Added option to get OpenCV source for version 4.5.1 from PyPI
 * Added option to use VTK 9.1.0 and made 9.1 the default
-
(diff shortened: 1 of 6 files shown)


daniel-riehm:
jenkins test this please

chetnieter:
It looks there has been several minor (4.4.2 & 4.4.3) and major (5.0.* and 5.1.*) release of FFMpeg since this PR was open. Should we consider upgrading ffmpeg if the fix has made it into one of these releases?

daniel-riehm:
@chetnieter Sure, we can look into that in another PR, but it would likely come with significant issues, as there are a number of existing patches we have for 4.4.1 that we would have to port over to 5.x. And moving to a new major version likely comes with some interface / behavior changes as well, which we would have to deal with in KWIVER. Are there any objections to merging this for the interim?

chetnieter (approved):

chetnieter:
> @chetnieter Sure, we can look into that in another PR, but it would likely come with significant issues, as there are a number of existing patches we have for 4.4.1 that we would have to port over to 5.x. And moving to a new major version likely comes with some interface / behavior changes as well, which we would have to deal with in KWIVER. Are there any objections to merging this for the interim?

Since we would have to bump up a major version patching it for now is fine.

PR #724, opened 2022-12-07
Allow use of x264 for older ffmpeg

mattdawkins:


daniel-riehm (approved):
Look good assuming CI passes.

PR #731, opened 2023-04-17
Target the Qt fix to only Makefiles

daniel-riehm:
Turns out that the Ninja generator does not like #729 (see [here](http://canifis:8080/view/Fletch/job/FletchLinuxNightly/38/console)), so this PR adds a check that we are using Makefiles before applying that fix.


daniel-riehm:
Mac failure unrelated

chetnieter (approved):

PR #737, opened 2023-09-29
Clean up Dockerfile

daniel-riehm:
This PR cleans up the `Dockerfile`, which has not been building properly for a long while. Notable changes include removal of the `PY_MAJOR_VERSION` argument (`python2` is long deprecated) and the neat `COPY --from=base / /` trick which reduces the size of the resulting container dramatically (`20GB -> 3GB`) by removing the layers describing the changes to intermediate build files, keeping only the installed final product.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.dockerignore
+dockerfile
--- a/dockerfile
+++ b/dockerfile
@@ -1,67 +1,94 @@
 # Fletch Dockerfile
 # Installs the fletch binary to /opt/kitware/fletch
 
-ARG BASE_IMAGE=ubuntu:18.04
-FROM ${BASE_IMAGE}
-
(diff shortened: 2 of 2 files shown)


chetnieter (approved):
Looks good to me. Do we know if this will impact the Kitware docker image?

https://github.com/Kitware/kwiver/blob/master/Dockerfile

If not we should check and prepare a MR for the KWIVER Dockerfile to insure it still builds.

daniel-riehm:
> Looks good to me. Do we know if this will impact the Kitware docker image?
> 
> https://github.com/Kitware/kwiver/blob/master/Dockerfile
> 
> If not we should check and prepare a MR for the KWIVER Dockerfile to insure it still builds.

Yes, I have a cleanup branch for KWIVER as well, but was waiting for this one to merge before creating a MR for it.

PR #722, opened 2022-12-05
Make a patched version of pybind11 2.10.3 available in Fletch

chetnieter:
PR to test if pybind11 version 2.9.2 will break KWIVER

PR #740, opened 2024-01-07
pybind11: update to 2.11.1

mathstuf:
This supports newer Python versions.


chetnieter:
There is a similar PR #722 that includes a patch needed for the kwiver v2 work.

mathstuf:
TODO: Adapt the 2.10.3 "patch"

PR #742, opened 2024-02-13
gdal: ensure that an rpath to the install tree is set

mathstuf:
Without this, loading GDAL was failing to find its dependencies (e.g., `libproj.so.18`) when not in a standard location.


mathstuf:
@chetnieter

PR #747, opened 2024-03-04
More tweaks to fletch, removing Qt4, non-free OpenCV, add GLog_DIR

aronhelser:
Remove Qt 4.x, it's much too old to be useful.

Turn on non-free components for OpenCV in 4.9.0, needed in the kwiver arrow.

Add correct GLog_DIR so glog can be found correctly by the ceres arrow.


aronhelser:
@mathstuf want to take a look?

mathstuf:
I can't believe that this is a single commit :) .

PR #749, opened 2024-03-14
Add eigen 3.4.0 as an option

ChristosT:
Eigen `3.4.0` is required for compiling kwiver with OpenCV 4.9.0 on windows.

The default value was left on `3.3.9`


CMake/fletch-tarballs.cmake
```diff
@@ -139,10 +139,21 @@ endif()
...
-set(Eigen_md5 "609286804b0f79be622ccf7f9ff2b660")
-set(Eigen_dlname "eigen-${Eigen_version}.tar.gz")
+set(Eigen_SELECT_VERSION 3.3.9 CACHE STRING "Select the version of Eigen to build.")
+set_property(CACHE Eigen_SELECT_VERSION PROPERTY STRINGS "3.3.9" "3.4.0")
+mark_as_advanced(Eigen_SELECT_VERSION)
+
+set(_Eigen_version ${Eigen_SELECT_VERSION})
+if(_Eigen_version VERSION_EQUAL 3.3.9)
+  set(Eigen_md5 "609286804b0f79be622ccf7f9ff2b660")
+elseif(_Eigen_version VERSION_EQUAL 3.3.4)
```
aronhelser, line 149: Your version here doesn't match. There's no case for 3.4.0

ChristosT: oups you are right !

PR #751, opened 2024-03-18
Clarify what building Fletch means

### LINKED ISSUE #750: Clarify what "Building Fletch" means
[Building Fletch](https://github.com/Kitware/fletch?tab=readme-ov-file#building-fletch) is a section in the README, but since it also says Fletch is “a pure CMake project” it's unclear what it might mean to build fletch, and the README is confusing.  I can, if I stretch my brain a bit, imagine that Fletch contains some generated CMake files, perhaps?  What Fletch actually _is_ would be a lot clearer if you'd spell that info out.
Thanks

Erotemic:
A minor README update that clarifies what "building fletch" means.

Resolves #750

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -96,6 +96,8 @@ or build the Fletch image using the dockerfile::
 Building Fletch
 ===============
 
+Building Fletch results in a set of inter-compatable shared libraries suitable for use in downstream applications - with the primary consumer being `KWIVER <https://github.com/Kitware/kwiver>`_. Using the CMake configuration, the user specifies which libraries to build, and Fletch handles ensuring the chosen set is compatible.
+
 Dependencies
 ------------
 
--- a/README.rst
+++ b/README.rst
@@ -96,7 +96,12 @@ or build the Fletch image using the dockerfile::
 Building Fletch
 ===============
 
-Building Fletch results in a set of inter-compatable shared libraries suitable for use in downstream applications - with the primary consumer being `KWIVER <https://github.com/Kitware/kwiver>`_. Using the CMake configuration, the user specifies which libraries to build, and Fletch handles ensuring the chosen set is compatible.
+Building Fletch results in a set of inter-compatable shared libraries suitable
+for use in downstream applications - with the primary consumer being
+`KWIVER <https://github.com/Kitware/kwiver>`_. Using the CMake configuration,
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -96,6 +96,8 @@ or build the Fletch image using the dockerfile::
 Building Fletch
 ===============
 
+Building Fletch results in a set of inter-compatable shared libraries suitable for use in downstream applications - with the primary consumer being `KWIVER <https://github.com/Kitware/kwiver>`_. Using the CMake configuration, the user specifies which libraries to build, and Fletch handles ensuring the chosen set is compatible.
+
```
aronhelser, line 100: This is a good clarification, but I don't think Fletch does anything about "ensuring the chosen set is compatible". They are all compiled with shared settings, like debug or release flags, but that's about it.

mleotta: I disagree.  Compatibility is an important thing that Fletch provides.  The various packages need to make compatible from a dependency graph sense.  The problem we often had without Fletch is that package A requires package C and package B also requires package C, but if A and B are build independently with default settings they will pull in different versions of package C that will conflict when you try to build something (e.g. KWIVER) on top of A and B.  There are other build configuration settings that pop up as well that prevent packages from working together if not all configured in a consistent way.

Erotemic: When I wrote "ensures", I felt that it was probably too strong. I do recall seeing code that flags when specific requested versions were incompatabile, which is where the thought came from. Open to suggestions on rewording.

aronhelser: Ah, ok, this compatibility is provided by the build setup. I recognize that this is what most of the code in Fletch is dedicated to - cmake code that provides the setup that works to build Kwiver on top of Fletch. 

The experience that I had that is made me raise my objection is that Fletch doesn't configure the dependency graph for you. When I enabled Ceres, I got the cmake error "Ceres requires SuiteSparse". I enabled SuiteSparse and got the error "SuiteSparse requires LAPACK". But once I got things turned on, with the right versions selected, everything built and worked together.

aronhelser: I would just replace "ensures" with "tries to ensure" and it sounds good to me.

Erotemic: pushed up the change.

aronhelser: +1, LGTM!

PR #752, opened 2024-03-21
Only write registration descriptor when async

daniel-riehm:
The standard only requires us to write the registration descriptor when the KLV stream is asynchronous; currently we always write it.


daniel-riehm:
jenkins test this please

PR #753, opened 2024-04-05
CMake variable tweak so VTK builds with Qt when Qt enabled.

aronhelser:


aronhelser:
Qt_FOUND is not set in my builds of fletch.

aronhelser:
@chetnieter

daniel-riehm:
@aronhelser @chetnieter This causes compilation errors in CI; see [here](https://open.cdash.org/viewBuildError.php?buildid=9556341) for example.

PR #754, opened 2024-04-16
Add option to omit writing SDT

daniel-riehm:
The [SDT](https://en.wikipedia.org/wiki/Service_Description_Table) is automatically inserted by FFmpeg whenever it writes a transport stream. This PR adds an option to disable that insertion, to opt out of the default service metadata (`service_name: Service01`, `service_provider: FFmpeg`) being added in. The SDT was designed for e.g. broadcast television and is likely useless in computer vision applications.

### DIFF
--- a/Patches/FFmpeg/4.4.1/mpegtsenc.c
+++ b/Patches/FFmpeg/4.4.1/mpegtsenc.c
@@ -114,6 +114,7 @@ typedef struct MpegTSWrite {
     int64_t sdt_period_us;
     int64_t last_pat_ts;
     int64_t last_sdt_ts;
+    int omit_sdt;
 
     int omit_video_pes_length;
 } MpegTSWrite;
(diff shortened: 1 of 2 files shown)


Patches/FFmpeg/4.4.1/mpegtsenc.c
```diff
@@ -114,6 +114,7 @@ typedef struct MpegTSWrite {
     int64_t sdt_period_us;
     int64_t last_pat_ts;
     int64_t last_sdt_ts;
+    int omit_sdt;
```
chetnieter, line 117: Is there a ffmpeg specific reason this can't be a `bool`?

Patches/FFmpeg/5.1.2/mpegtsenc.c
```diff
@@ -120,6 +120,7 @@ typedef struct MpegTSWrite {
     int64_t last_pat_ts;
     int64_t last_sdt_ts;
     int64_t last_nit_ts;
+    int omit_sdt;
```
chetnieter, line 123: Can this be a `bool`?

chetnieter (changes requested):
Looks good except for the use of `int` where I would think a `bool` would be appropriate. Is there a FFMpeg reason for this?

Patches/FFmpeg/4.4.1/mpegtsenc.c
daniel-riehm: FFmpeg seems to prefer `int`s over `bool`s (see the next member here: `int omit_video_pes_length`). Could be for memory alignment, or to leave space for future expansion of options, or because FFmpeg is old and C89 didn't have a native `bool` type. Whatever the reason, I figured it's best to stick with the surrounding coding style.

Patches/FFmpeg/5.1.2/mpegtsenc.c
daniel-riehm: (See other comment)

chetnieter (approved):

PR #758, opened 2024-05-03
Update package defaults to reflect needs for kwiver-v2

chetnieter:
Update some of the defaults for packages so the defaults match what is currently used by the upcoming kwiver version 2 release.


aronhelser:
+1, assuming we're ignoring CI for now, and will update so it builds V2 in another MR

daniel-riehm:
@chetnieter Could you resolve the conflict and then we can merge this?

PR #760, opened 2024-06-05
Update cppzmq

johnkit:
Updates cppzmq to add option for a later release (4.10.0), which includes many API improvements. The previous version (4.2.3) is still the default.

Note: this supersedes #738
closes #738

### DIFF
--- a/CMake/External_cppzmq.cmake
+++ b/CMake/External_cppzmq.cmake
@@ -7,20 +7,34 @@ if(NOT fletch_ENABLE_ZeroMQ)
   message(WARNING "cppzmq module enabled without ZeroMQ! Only the header will be installed!")
 endif()
 
-ExternalProject_Add(cppzmq
-  URL ${cppzmq_url}
-  URL_MD5 ${cppzmq_md5}
-  ${COMMON_EP_ARGS}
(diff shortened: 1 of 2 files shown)


johnkit:
@dstoup Can you review this PR please? I don't seem to have permission to add you (or anyone) to the "Reviewers" section; so I would appreciate if you would also do that.

chetnieter:
@johnkit could we get this PR modified so it adds the new version of cppzmq as an option rather than replacing the old version. You can look at what is done for OpenCV to see how fletch supports multiple versions. I am not sure how this will impact the KWIVER 2.0 work and because things are in flux it is hard to test it right now. Once we determine that the new version of cppzmq works with the KWIVER 2.0 changes we can drop the older version, but for now I would like to keep it.

johnkit:
@chetnieter it appears that mac build failed and windows build still pending (?). Can you get more details from Jenkins?

chetnieter:
> @chetnieter it appears that mac build failed and windows build still pending (?). Can you get more details from Jenkins?

I am not worried about the Mac build. We have effectively dropped support for that awhile. Once the windows build run I will mrege.

chetnieter:
jenkins test this please

PR #763, opened 2024-08-01
Fix opencv python includes and numpy version in fletch

medmssaouri:
Goal : 
  - Fix bugs to generate release

Changes : 
- Fix:
   - OpenCV does not find python include files (ex: Python.h), because they are in ${fletch_BUILD_INSTALL_PREFIX}/include/ 
      python${CPython_version} and not  ${fletch_BUILD_INSTALL_PREFIX}/include
  - Set version of numpy according to version of python

### DIFF
--- a/CMake/External_CPython.cmake
+++ b/CMake/External_CPython.cmake
@@ -52,7 +52,7 @@ if( WIN32 )
       -DPYTHON_BASEPATH:PATH=${PYTHON_BASEPATH}
       -P ${fletch_SOURCE_DIR}/Patches/CPython/install_python_windows.cmake
   )
-  
+
   set( LIBNAME python${CPython_version_major}.lib )
 
   set( BUILT_PYTHON_EXE     ${BUILT_PYTHON_EXE}.exe )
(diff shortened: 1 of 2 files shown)


CMake/External_CPython.cmake
```diff
@@ -149,7 +149,13 @@ set( PYTHON_LIBRARY_DEBUG ${PYTHON_LIBRARY_DEBUG} )
...
 
 set( fletch_PYTHON_LIBS numpy cython ordered_set )
-set( fletch_PYTHON_LIB_CMDS "numpy==1.19.3" "Cython" "ordered_set" )
+set( fletch_PYTHON_LIB_CMDS "Cython" "ordered_set" )
+
+if( CPython_version VERSION_GREATER_EQUAL "3.8" )
+  list( APPEND fletch_PYTHON_LIB_CMDS "numpy==1.23.5")
+else()
+  list( APPEND fletch_PYTHON_LIB_CMDS "numpy==1.19.5")
+endif()
```
mattdawkins, line 158: We might be able to parametize these later (or just not have fletch install numpy by default)

PR #767, opened 2025-02-26
Make 2.10.3 the default pybind11 version

daniel-riehm:
This is the KWIVER 2.0-compatible version, which is what people should use going forward.


daniel-riehm:
Oops, duplicate of #758

PR #770, opened 2025-07-16
Now checking for PROJ when configuring GDAL version 3.11

chetnieter:
When I updated GDAL to support version 3.11 I did not add the needed logic to insure PROJ is available and is a dependency for GDAL. This is because GDAL switched to cmake at version 3.5.


mleotta (approved):
I didn't test it, but it looks okay to me

PR #781, opened 2026-01-08
Updates for C++17 compilation required on some systems linking against kwiver master

mattdawkins:


CMake/External_OpenCV.cmake
```diff
@@ -312,7 +312,7 @@ if (CMAKE_COMPILER_IS_GNUCC AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5)
 endif()
 
 # OpenCV 3.3.0 has an option to enable C++ 11
-if (fletch_BUILD_CXX11)
+if (fletch_BUILD_CXX17)
   list(APPEND OpenCV_EXTRA_BUILD_FLAGS -DENABLE_CXX11:BOOL=ON)
```
daniel-riehm: Shouldn't `ENABLE_CXX11` be unconditionally enabled now that presumably that is the baseline in fletch? And/or does this need line to be changed to `ENABLE_CXX17`?

daniel-riehm (commented):
I think it makes sense to have C++11 be the standard now that C++17 has superseded it, instead of the default still being C++98. Unless this causes a problem?

